PR #1, opened 2016-05-16
add applications.active boolean flag

pkarman:
Rationale:
Service providers may want to register applications, and then toggle them on/off.


sbc100:
lgtm

amoose:
👌 I think you are the only one with write-access to this repo @pkarman so you will have to merge :P

PR #2, opened 2016-05-17
Admin users

pkarman:
This PR does a few things:

* adds the `admin` role to the `User` model
* adds authorization via Pundit to the `Application` model (https://github.com/18F/identity-private/issues/361)
* adds `approved` flag to the `Application` model (https://github.com/18F/identity-private/issues/357)
* adds email notification when `Application` is created and approved

**Why**

Until now the dashboard has had no authorization checks at all. This PR implements a human intervention step into the activation of new `Application` instances, and verifies that users may not interact with each others' applications, unless the user is an admin.


app/controllers/users/applications_controller.rb
```diff
@@ -26,9 +29,11 @@ def new
     end
 
     def edit
+      authorize application
```
sashax: I'm curious why these aren't before_actions. (There may be a totally legit reasons, of course, but I understood that as the Ruby Way)

app/controllers/users/applications_controller.rb
```diff
@@ -37,23 +42,58 @@ def application
       @application ||= Application.find_by(issuer: params[:id])
     end
 
+    def authorize_approval
+      if params.require(:application).key?(:approved) && !current_user.admin?
+        raise Pundit::NotAuthorizedError, I18n.t('dashboard.errors.not_authorized')
+      end
+    end
+
```
amoose: before_filter for DRY-ness?

spec/features/applications_spec.rb
```diff
@@ -41,14 +45,71 @@
...
+  scenario 'admin attempts to view' do
+    admin_user = create(:user, admin: true)
+    app = create(:application)
+    login_as(admin_user)
+
+    visit users_application_path(app)
+
+    expect(page.status_code).to eq(200)
+  end
 end
```
amoose: these would be more appropriate in a request spec for performance

spec/features/applications_spec.rb
```diff
@@ -41,14 +45,71 @@
...
-    
+
+feature 'Admin User Approval' do
+  scenario 'new application' do
+    ActionMailer::Base.deliveries.clear
```
amoose: create a [helper](https://github.com/18F/identity-idp/blob/master/spec/support/features/mailer_helper.rb) for clearing deliveries.

spec/features/applications_spec.rb
```diff
@@ -41,14 +45,71 @@
...
+
+    choose('application_approved_true', option: 'true')
+    click_on 'Update'
+
+    expect(page).to have_content('Success')
+    app.reload
+    expect(app.approved?).to eq true
+    expect(ActionMailer::Base.deliveries.count).to eq(4)
+  end
+end
```
amoose: also break out in to individual specs

spec/policies/application_policy_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      expect(ApplicationPolicy).to_not permit(other_user, app)
+    end
+  end
+
+  permissions :show? do
+    it 'allows owner or admin to show' do
+      user = build(:user)
+      other_user = build(:user)
+      admin_user = build(:user, admin: true)
+      app = build(:application, user: user)
```
amoose: create a single group of `let`s for each type of user

spec/requests/applications_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+    it 'disallows app owner from approving the app' do
+      app = create(:application)
+      login_as(app.user)
+
+      put users_application_path(app), { application: { approved: 'true' } }
+
+      expect(response.status).to eq(401)
+    end
+  end
+end
```
amoose: add spec for non-owner

amoose:
Nice work, PK.. noticed a few areas to DRY up or room to enhance specs.

app/controllers/users/applications_controller.rb
pkarman: It is the Rails Way, definitely.

In this case I thought it was clearer to put the method calls exactly they are needed, since about half the actions in that controller do not need the `before_action` (i.e. the whitelist is as long as the blacklist).

I will change for Rails Consistency though.

app/controllers/users/applications_controller.rb
pkarman: it's only called in one action, but I will change for consistency.

pkarman:
@amoose @sashax thanks for the reviews. I think I've hit all your comments.

Per slack conversation, should I be squashing my commits on PRs? I typically do not, since I like to see the messy history of things, but I understand why the `master` history might be preferable as a "clean" commit-per-PR.

amoose:
This is your baby! 🚼  if you want verbose a verbose commit log, you got it!

amoose:
Nice work PK! 🍷

PR #7, opened 2016-06-10
Add Code Climate and linter configs

monfresh:
**Why**: To ensure consistency in our code style and quality

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,30 @@
+engines:
+  brakeman:
+    enabled: true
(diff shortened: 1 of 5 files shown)


monfresh:
Note that there are a bunch of issues in this repo reported by Code Climate, but since they weren't introduced in this PR, they shouldn't be addressed in this PR. I think the vast majority are stylistic and can be autocorrected by running `rubocop -a`, which I can do in a separate PR.

pkarman:
lovely thanks

PR #8, opened 2016-06-10
run rubocop -a

pkarman:
This should clean up all the rubocop infractions that can be cleaned up automatically.

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -8,7 +8,7 @@ Devise.setup do |config|
   # confirmation, reset password and unlock tokens in the database.
   # Devise will use the `secret_key_base` as its `secret_key`
   # by default. You can change it below and use your own secret key.
-  #config.secret_key = 'c9aac5e1283cb3302f0a7a2217dbd287a8b8564070b519d96246666ba172191d3b1267d1b578c0e1771890a15fe2ef30edbd3adc3cac38d6b9a2b918b8421ab9'
+  # config.secret_key = 'c9aac5e1283cb3302f0a7a2217dbd287a8b8564070b519d96246666ba172191d3b1267d1b578c0e1771890a15fe2ef30edbd3adc3cac38d6b9a2b918b8421ab9'
 
   # ==> Mailer Configuration
   # Configure the e-mail address which will be shown in Devise::Mailer,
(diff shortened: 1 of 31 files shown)


config/initializers/devise.rb
```diff
@@ -8,7 +8,7 @@
   # confirmation, reset password and unlock tokens in the database.
   # Devise will use the `secret_key_base` as its `secret_key`
   # by default. You can change it below and use your own secret key.
-  #config.secret_key = 'c9aac5e1283cb3302f0a7a2217dbd287a8b8564070b519d96246666ba172191d3b1267d1b578c0e1771890a15fe2ef30edbd3adc3cac38d6b9a2b918b8421ab9'
+  # config.secret_key = 'c9aac5e1283cb3302f0a7a2217dbd287a8b8564070b519d96246666ba172191d3b1267d1b578c0e1771890a15fe2ef30edbd3adc3cac38d6b9a2b918b8421ab9'
```
sbc100: I would leave this file as-is.  The comments style is designed to distinguish between commented out settings and docs about the settings.  Rubocop doesn't seem to grok this.

sbc100:
lgtm % comments

pkarman:
thanks @sbc100 -- devise.rb config reverted.

config/initializers/devise.rb
monfresh: Rubocop doesn't seem to grok this intentionally. I don't see any configuration that would allow a comment in this scenario to not have a leading space. I also don't see any GitHub issues complaining about this, which perhaps indicates that people agree in general with leading spaces before comments everywhere.

For example, the default config files generated by Rails adhere to the leading space before comments everywhere. Even the Devise config where this offense was flagged has leading spaces everywhere else, so the lack of space on this line is an anomaly with respect to the rest of the file. On their master branch, even the secret_key has a leading space, so either this was a copy and paste error, or it used to be a bug in Devise which has now been fixed. See: https://github.com/plataformatec/devise/blob/0c1b1a896f549f92d76baf06fcece150420e94bf/lib/generators/templates/devise.rb

Hence, I vote to fix this.

monfresh:
@pkarman @sbc100 Looks like my comment got deleted after your second commit. Perhaps you might have received it via an email notification. If not, I will summarize. I don't agree with Sam's comment. The Ruby Style Guide does not differentiate between different types of comments. You either follow the style or you don't. The point is to be consistent everywhere with comments.

The default Devise config file follows the style guide, as you can see here: https://github.com/plataformatec/devise/blob/0c1b1a896f549f92d76baf06fcece150420e94bf/lib/generators/templates/devise.rb

The fact that the `secret_key` line didn't have a leading space was most likely a copy and paste error because the template that Devise uses to generate the file has always used leading spaces.

pkarman:
@monfresh I did get the email notification.

I have zero feelings about this issue one way or the other. My only preference is to minimize rubocop complaints.

monfresh:
Sounds good. Then the way to minimize Rubocop complaints is to revert back to your original commit that fixed the missing leading space.

pkarman:
+1 done.

monfresh:
👍

PR #10, opened 2016-06-13
fix email subject for application create/approve notifications

pkarman:
Fixes bug in email subject where I18n key was wrong, and adds tests to verify.

### DIFF
--- a/spec/requests/applications_spec.rb
+++ b/spec/requests/applications_spec.rb
@@ -22,15 +22,53 @@ describe 'Users::Applications' do
     end
 
     it 'allows admin to approve' do
-      app = create(:application)
-      admin_user = create(:user, admin: true)
-      login_as(admin_user)
+      ClimateControl.modify ADMIN_EMAIL: 'identity-admin@example.com' do
+        app = create(:application)
+        admin_user = create(:user, admin: true)
+        deliveries.clear
+        login_as(admin_user)
 
-      put users_application_path(app), application: { approved: 'true' }
+        mailer = instance_double(ActionMailer::MessageDelivery)
+        allow(UserMailer).to receive(:admin_approved_application).with(anything).and_return(mailer)
+        allow(UserMailer).to receive(:user_approved_application).with(anything).and_return(mailer)
+        allow(mailer).to receive(:deliver_later)
 
-      expect(response.status).to eq(302) # redirect on success
-      app.reload
-      expect(app.approved?).to eq(true)
+        put users_application_path(app), application: { approved: 'true' }
+
+        expect(response.status).to eq(302) # redirect on success
+        app.reload
+        expect(app.approved?).to eq(true)
+
+        expect(UserMailer).to have_received(:admin_approved_application).with(app)
+        expect(UserMailer).to have_received(:user_approved_application).with(app)
+        expect(mailer).to have_received(:deliver_later).twice
+      end
+    end
+  end
+
+  describe 'notifications' do
+    it 'sends email to admin requesting approval' do
+      ClimateControl.modify ADMIN_EMAIL: 'identity-admin@example.com' do
+        user = create(:user)
+        deliveries.clear
+        login_as(user)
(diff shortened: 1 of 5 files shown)


spec/requests/applications_spec.rb
```diff
@@ -24,13 +24,35 @@
...
+      login_as(user)
+
+      post users_applications_path, application: { name: 'test' }
+
+      app = Application.last
+      expect(response.status).to eq(302)
+      expect(deliveries.count).to eq(2)
+      expect(deliveries[0].subject).to eq(
+        I18n.t('dashboard.mailer.new_application.subject', id: app.issuer)
+      )
```
monfresh: Should we also add an expectation that verifies that the email is sent to the admin (as opposed to the user, for example)?

spec/requests/applications_spec.rb
```diff
@@ -24,13 +24,35 @@
...
 
       put users_application_path(app), application: { approved: 'true' }
 
       expect(response.status).to eq(302) # redirect on success
       app.reload
       expect(app.approved?).to eq(true)
+      expect(deliveries.count).to eq(2)
+      expect(deliveries[0].subject).to eq(
+        I18n.t('dashboard.mailer.approved_application.subject', id: app.issuer)
+      )
```
monfresh: Same comment here about verifying the recipient.

pkarman:
good call, @monfresh. I had to think a bit about the best way to test that, since the emails are not in a predictable order. See if this latest commit helps.

monfresh:
Hmm. I left a comment here yesterday about only testing that the appropriate UserMailer is called in the request specs, and then adding mailers specs to test the actual mailer, but I don't see it anymore. Did you receive it?

pkarman:
@monfresh I did not see that comment. Odd.

I could restructure the tests that way, which would necessitate refactoring where/how the UserMailer is used in the controller. Happy to do that.

monfresh:
You don't need to change your controller. Here is how you can test that the correct mailers are being called:
```ruby
describe 'notifications' do
  it 'sends email to admin requesting approval' do
    ClimateControl.modify ADMIN_EMAIL: 'identity-admin@example.com' do
      user = create(:user)
      deliveries.clear
      login_as(user)

      mailer = instance_double(ActionMailer::MessageDelivery)
      allow(UserMailer).to receive(:admin_new_application).with(anything).and_return(mailer)
      allow(UserMailer).to receive(:user_new_application).with(anything).and_return(mailer)
      allow(mailer).to receive(:deliver_later)

      post users_applications_path, application: { name: 'test' }

      expect(UserMailer).to have_received(:admin_new_application).with(Application.last)
      expect(UserMailer).to have_received(:user_new_application).with(Application.last)
      expect(mailer).to have_received(:deliver_later).twice

      expect(response.status).to eq(302)
    end
  end
end
```

monfresh:
Then your mailer specs would look something like this: https://github.com/18F/identity-idp/blob/master/spec/mailers/user_mailer_spec.rb

pkarman:
@monfresh thanks for the detailed examples. I have committed those changes.

spec/requests/applications_spec.rb
```diff
@@ -22,15 +22,52 @@
...
+        app = Application.last
+
+        expect(UserMailer).to have_received(:admin_new_application).with(app)
+        expect(UserMailer).to have_received(:user_new_application).with(app)
+        expect(mailer).to have_received(:deliver_later).twice
 
-      expect(response.status).to eq(302) # redirect on success
-      app.reload
-      expect(app.approved?).to eq(true)
+        expect(response.status).to eq(302)
```
monfresh: Looks great! Perhaps add another expectation here to make sure the app is not yet approved?

pkarman:
@monfresh done

monfresh:
Looks good. Wanna squash?

pkarman:
:squash:'d

PR #12, opened 2016-06-14
Set up app to report test coverage to Code Climate

monfresh:
**Why**: To make it easy to track our test coverage and code quality all in one place.


monfresh:
Ready to review now. I worked out all the kinks with Code Climate. To verify, Travis should report that it successfully submitted test coverage to Code Climate.

pkarman:
+1 thanks

PR #13, opened 2016-06-24
Add tests for SSO/SLO with fake IdP

pkarman:
**Why**:

Confirm the SSO/SLO workflows with our own fake IdP server.


monfresh:
LGTM.

PR #15, opened 2016-06-27
use IdP cert fingerprint instead of full cert

pkarman:


lib/saml_config.rb
```diff
@@ -10,32 +10,37 @@ def settings
...
+        double_quote_xml_attribute_values: true,
+        security: {
+          authn_requests_signed: true,
+          embed_sign: true,
+          digest_method: 'http://www.w3.org/2001/04/xmlenc#sha256',
+          signature_method: 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256'
+        }
+      )
     end
+    # rubocop:enable AbcSize, MethodLength
```
monfresh: To make Rubocop happy, you could store this configuration in a constant, like here: https://github.com/18F/identity-idp/blob/master/config/initializers/omniauth.rb

monfresh: Also, why Hashie::Mash?

pkarman: Hashie::Mash is so the resulting object "acts like" a RubySaml::OneLogin::Config object (i.e. it has methods matching hash keys).

pkarman: I looked at doing it as a constant, it just seemed tricky because I was pulling default values from `saml_env_config` which meant I needed to load that file, and defer it till runtime.

monfresh: Gotcha. Sounds good.

PR #16, opened 2016-07-07
Fix IdP fingerprint for SAML feature tests

pkarman:
**Why**:

7f8e331936d799e7d7fc1c872dbe848acecb47d0 changed the config
to use a cert fingerprint instead of a raw cert. Need to also
change the test idp_fingerprint to be the fingerprint
of the local SP cert, since the SP cert is used by FakeSamlIdp server.


monfresh:
👍

PR #19, opened 2016-07-07
Whitelist all IdP domains for Content Security Policy

pkarman:
Equivalent to what the -idp app does for all service providers.


amoose:
LGTM! 👍

PR #20, opened 2016-07-08
Add capistrano and move to login.gov domain

jgrevich:


jgrevich:
This should be good to go once #26 is pulled in.

amoose:
GTG 🚢 
We will finalize deployment environments shortly between: `tf qa dev demo`

PR #21, opened 2016-07-11
Cloud dot gov fixes

pkarman:
Several small changes to how app is deployed to cloud.gov, most importantly turning off `embed_sign` so that redirect URLs are smaller than 4k.


config/initializers/secure_headers.rb
```diff
@@ -11,7 +17,7 @@
     frame_src: %w('self'), # deprecated in CSP 2.0
     child_src: %w('self'), # CSP 2.0 only; replaces frame_src
     # frame_ancestors: %w('self'), # CSP 2.0 only; overriden by x_frame_options in some browsers
-    form_action: %w('self'), # CSP 2.0 only
+    form_action: %w('self') + whitelisted_domains.uniq!, # CSP 2.0 only
```
monfresh: The `secure_headers` gem already removes duplicates from this list, so I wouldn't duplicate the effort.

monfresh: Also, do the whitelisted domains apply to all controller actions in this app, or only when signing in and out?

pkarman: currently it's all controller actions, because it's the logout link in the nav which appears on every page.

However I am currently working on refactor to switch to SP-initiated SLO which may mean this isn't required at all.

pkarman:
@monfresh I made `/users/logout` a GET accessible route, so no need to whitelist any domains for a pseudo form button/link.

config/routes.rb
```diff
@@ -10,6 +10,7 @@
     post '/users/sessions' => 'users/sessions#create', as: :user_session
     delete '/users/sessions' => 'users/sessions#destroy', as: :destroy_user_session
```
monfresh: should this be removed if we're going to use `get` for logging out?

monfresh:
This looks good. Are there any security concerns with using `GET`, or issues with bots triggering the sign out link?

pkarman:
Session management is with Devise, so an un-authenticated GET request to logout would either be denied or a no-op. Typically I don't like to make a non-idempotent request into a GET; this is just a convenience to avoid the browser security. To be honest, I'm not sure exactly sure why DELETE or POST poses a problem, since the action is `self` which then just re-directs.

Moving to SP-initiated logout in #23 may change this yet again, so I will revisit once both these PRs are merged.

pkarman:
Please merge this BEFORE #23 as it will cause a needed merge conflict with #23 that I can more cleanly resolve.

monfresh:
👍

PR #22, opened 2016-07-11
Rename Application model to ServiceProvider

pkarman:
Also zaps lots of unused i18n strings.

**Why**:

Naming consistency across apps


app/views/user_mailer/user_new_service_provider.html.slim
```diff
@@ -3,10 +3,10 @@ html
     meta content='text/html; charset=UTF-8' http-equiv='Content-Type'
   body
 
-  p You have created a new Identity service provider application:
+  p You have created a new Identity service provider service_provider:
```
amoose: Too many service providers!

config/routes.rb
```diff
@@ -15,7 +15,7 @@
   end
 
   namespace :users do
-    resources :applications do
+    resources :service_providers do
     end
   end
```
amoose: I don't think we need to address this any time soon, but note that service provider entries will need to have delegated access at some point.

amoose:
Nice work. 👌

config/routes.rb
pkarman: good point. My intention was to eventually add `Group` or `Org` in order to define authz for a group of `User`s.

app/views/user_mailer/user_new_service_provider.html.slim
pkarman: nice catch. search+replace FAIL. fixed now.

pkarman:
@amoose thanks for the review. Think I've hit all your comments.

PR #23, opened 2016-07-13
SP-initiated logout

pkarman:
**Why**:

Better UX, since we end up on dashboard site, where user
clicked the Sign Out button.

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -22,10 +22,9 @@ module Users
 
     def logout
       if params[:SAMLRequest]
-        saml_req = params[:SAMLRequest]
-        saml_req_decoded = Base64.decode64(saml_req)
-        Rails.logger.info saml_req_decoded
         idp_logout_request
+      elsif params[:SAMLResponse]
+        validate_slo_response
       else
         flash[:alert] = 'SAMLRequest missing'
         render :failure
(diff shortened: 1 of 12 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -34,31 +33,56 @@ def logout
...
 
+    def validate_slo_response
+      slo_response = OneLogin::RubySaml::Logoutresponse.new(params[:SAMLResponse], saml_settings)
+      if slo_response.validate
+        flash[:notice] = t('omniauth.logout_ok')
+        redirect_to root_url
+      else
+        flash[:alert] = t('omniauth.logout_fail')
+        render :failure
+      end
```
monfresh: Looks like this is not tested. By the way, I can tell because I'm using Code Climate's awesome browser extension, which you have to sign up to get, but I got access pretty quickly. https://codeclimate.com/browser-extension

monfresh: To clarify, what's not tested is the failure part of this method.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -34,31 +33,56 @@ def logout
...
-          nil,
-          RelayState: params[:RelayState]
-        )
-        redirect_to logout_response
+        redirect_to_idp_logout(logout_request)
       else
-        error_msg = "IdP initiated LogoutRequest was not valid: #{logout_request.errors}"
-        Rails.logger.error error_msg
-        render inline: error_msg
+        invalid_logout_request
```
monfresh: Add a test for this please.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -34,31 +33,56 @@ def logout
...
+        nil,
+        RelayState: params[:RelayState]
+      )
+      redirect_to logout_response
+    end
+
+    def invalid_logout_request
+      error_msg = "IdP initiated LogoutRequest was not valid: #{logout_request.errors}"
+      Rails.logger.error error_msg
+      render inline: error_msg
```
monfresh: Add a test for this please.

pkarman:
@monfresh thanks for the review. I've added tests for the failure flows.

app/controllers/users/omniauth_callbacks_controller.rb
pkarman: thanks, failure tests added.

db/migrate/20160715161608_uui_dto_string.rb
```diff
@@ -0,0 +1,5 @@
+class UuiDtoString < ActiveRecord::Migration
+  def change
+    change_column :users, :uuid, :string, null: false
```
pkarman: since the -idp seed users have UUID strings that are not actually valid UUIDs (not strictly base64-encoded), use a more generic `string` definition. As long as it is a MBUN we don't really care.

pkarman:
tests here will fail until https://github.com/sportngin/saml_idp/pull/50 is merged.

pkarman:
https://github.com/sportngin/saml_idp/pull/50 appears to have been abandoned. I have switched this PR to point at the 18F fork.

pkarman:
@amoose now that https://github.com/18F/saml_idp/pull/3 is merged I have updated this PR to point at master.

amoose:
LGTM 👍

spec/requests/slo_spec.rb
```diff
@@ -1,15 +1,67 @@
...
+      idp_uri = URI(response.headers['Location'])
+      saml_idp_resp = Net::HTTP.get(idp_uri)
+
+      # send the SAMLResponse back to our SP
+      post '/users/auth/saml/logout', SAMLResponse: saml_idp_resp
+
+      # expect we are logged out, on our site
+      expect(response).to redirect_to(root_url)
+      expect(flash[:notice]).to eq I18n.t('omniauth.logout_ok')
+    end
```
amoose: The only gap I see is IdP-initiated logout.

PR #26, opened 2016-08-02
Copy rubocop custom config from IdP

jgrevich:
__Why__
Rubocop was throwing errors at the Capistrano DSL with my capify branch. We have
already configured this to work correctly on the IdP hence we'll copy that
config.

__How__
Copy the config from the IdP app.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,12 +1,19 @@
+# This configuration only includes the cops that differ from the Rubocop
+# defaults, which can be found here:
+# https://github.com/bbatsov/rubocop/blob/master/config/default.yml
+# https://github.com/bbatsov/rubocop/blob/master/config/enabled.yml
+# https://github.com/bbatsov/rubocop/blob/master/config/disabled.yml
+
 AllCops:
   Include:
     - '**/Gemfile'
     - '**/Rakefile'
   Exclude:
-    - 'bin/**/*'
     - 'db/schema.rb'
-    - 'spec/i18n_spec.rb'
+    - 'bin/**/*'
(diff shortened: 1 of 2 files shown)


.rubocop.yml
```diff
@@ -1,12 +1,19 @@
...
+# https://github.com/bbatsov/rubocop/blob/master/config/disabled.yml
+
 AllCops:
   Include:
     - '**/Gemfile'
     - '**/Rakefile'
   Exclude:
-    - 'bin/**/*'
     - 'db/schema.rb'
-    - 'spec/i18n_spec.rb'
```
jgrevich: We should probably keep this right?

.rubocop.yml
```diff
@@ -24,6 +31,8 @@ Metrics/ClassLength:
   Max: 100
   Exclude:
   - spec/**/*
+  - app/controllers/users/confirmations_controller.rb
+  - app/controllers/devise/two_factor_authentication_controller.rb
```
jgrevich: 🔫

amoose:
GTG % minor code climate issues

.rubocop.yml
monfresh: No, the offenses reported for that file are valid. We can fix them in a separate PR.

monfresh: Or, you can run `rubocop -a spec/i18n_spec.rb` to auto fix it and check it in here if you want

.rubocop.yml
monfresh: Yes, please remove these.

PR #31, opened 2016-11-03
Fix production secrets

zachmargolis:
**Why**:
I thought default was shared across all environments, it is shared
across all enviromments _except_ production, so this brings back
the secrets I removed from prod accidentally in 4d7a70f9498367

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -14,3 +14,5 @@ staging:
 
 production:
   secret_key_base: <%= ENV['SECRET_KEY_BASE'] %>
+  http_auth_username: <%= ENV["SP_NAME"] %>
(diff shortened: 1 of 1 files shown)


jessieay:
> I thought default was shared across all environments, it is shared across all enviromments except production

I did not know this! good to know

jgrevich:
Are we sure? I think this is related to it not picking up env variables from `.bash_profile` since capistrano is running from a non-login, non-interactive shell http://capistranorb.com/documentation/faq/why-does-something-work-in-my-ssh-session-but-not-in-capistrano/

zachmargolis:
@jgrevich @jessieay It's just how this app was set up, we didn't put a `<<: *default` splat under `production` so we just aren't set up to share those things

PR #32, opened 2016-11-03
Copy over passenger restart steps

zachmargolis:
**Why**:
We want passenger to restart after each deploy

cc @jgrevich 

see: https://github.com/18F/identity-idp/pull/538/files


jgrevich:
👍

PR #33, opened 2016-11-04
Write deploy.json file

zachmargolis:
**Why**:
so we can expose the status of this deploy in our dashboard

(this is different from the other PR in that I'm avoiding all the CORS shenanigans since the current implementation is server-side and does not need to worry about CORS, we can just rely 100% on static files being served)

Based on https://github.com/18F/identity-idp/pull/663 -- would love a better way to share code for this, but it's extremely simple

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -25,4 +25,28 @@ set :tmp_dir, '/srv/idp'
 # TASKS
 #########
 namespace :deploy do
+  desc 'Write deploy information to deploy.json'
+  task :deploy_json do
+    on roles(:web), in: :parallel do
+      require 'stringio'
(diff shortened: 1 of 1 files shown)


config/deploy.rb
```diff
@@ -25,4 +25,28 @@
 # TASKS
 #########
 namespace :deploy do
+  desc 'Write deploy information to deploy.json'
+  task :deploy_json do
+    on roles(:app, :web), in: :parallel do
```
amoose: at second glance, shouldn't this only occur on the :web instance?

zachmargolis: 👍

PR #34, opened 2016-11-04
Prettify deploy dashboard, update our global stylez

brendansudol:
preview:
screen shot 2016-11-04 at 11 57 56 am


app/services/deploy_status_checker.rb
```diff
@@ -1,22 +1,39 @@
...
+  Deploy = Struct.new(:app, :env, :host)
+
+  Environment = Struct.new(:env, :deploys)
 
   DEPLOYS = [
-    Deploy.new('prod'),
-    Deploy.new('demo', 'idp.demo.login.gov'),
-    Deploy.new('dev', 'idp.dev.login.gov'),
-    Deploy.new('qa', 'idp.qa.login.gov')
+    Deploy.new('identity-idp', 'prod'),
```
zachmargolis: note to future selves, we probably want to start pulling these from a config or something so we don't have to hardcode in the source code

pkarman:
so pretty!

brendansudol:
i live to serve 🇺🇸

PR #37, opened 2016-11-04
Fix hosts for demo env

zachmargolis:
**Why**: too much copy-pasta

I am very open to how we can make these config-based, like parse a JSON from an ENV key? Anybody have any suggestions? @pkarman @jgrevich @amoose ?


pkarman:
I think it's fine as is. We aren't going to see (or want) more envs to check, once these are correctly configured in code, and changing them should require PR review anyway, imo.

PR #44, opened 2017-01-04
Adds 2 second open_timeout to HTTP request

amoose:
### Why
When checking the deploy status on our applications, we must
open an HTTP connection to the API; when the service is completely
down, the Rack::Timeout is reached before the individual API calls
time out.

### How
Add an open_timeout to the HTTP request to each individual API.


jgrevich:
LGTM! 👍

PR #45, opened 2017-01-13
Add /api/service_providers endpoint

pkarman:
**Why**: Allows active, approved Service Providers to be published
for consumption by identity-idp

### DIFF
--- a/app/serializers/service_provider_serializer.rb
+++ b/app/serializers/service_provider_serializer.rb
@@ -1,0 +1,44 @@
+class ServiceProviderSerializer < ActiveModel::Serializer
+  attributes(
+    :issuer,
+    :agency,
+    :friendly_name,
+    :metadata_url,
+    :acs_url,
+    :assertion_consumer_logout_service_url,
+    :cert,
+    :block_encryption,
+    #:redirect_url,
+    #:return_to_sp_url,
+    #:logo,
+    #:attribute_bundle,
+    :updated_at,
+    :signature
+  )
+
+  def agency
+    object.name
+  end
+
+  def friendly_name
+    object.description
+  end
+
+  def cert
+    object.saml_client_cert
+  end
+
+  def updated_at
+    object.updated_at.iso8601
+  end
+
+  def signature
+    Digest::SHA256.hexdigest unique_identifier
+  end
+
+  private
+
+  def unique_identifier
+    object.id.to_s + object.issuer + object.created_at.to_s + object.updated_at.to_s
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -14,10 +14,18 @@ Rails.application.routes.draw do
     get 'timeout' => 'users/sessions#timeout'
   end
 
-  namespace :users do
-    resources :service_providers do
-    end
-  end
+  get '/users/service_providers' => 'users/service_providers#index', as: 'users_service_providers'
+  post '/users/service_providers' => 'users/service_providers#create'
+  get '/users/service_providers/new' => 'users/service_providers#new',
+      as: 'new_users_service_provider'
+  get '/users/service_providers/:id/edit' => 'users/service_providers#edit',
+      as: 'edit_users_service_provider'
+  get '/users/service_providers/:id' => 'users/service_providers#show', as: 'users_service_provider'
+  patch '/users/service_providers/:id' => 'users/service_providers#update'
+  put '/users/service_providers/:id' => 'users/service_providers#update'
+  delete '/users/service_providers/:id' => 'users/service_providers#destroy'
+
+  get '/api/service_providers' => 'api/service_providers#index'
 
   root to: 'home#index'
 end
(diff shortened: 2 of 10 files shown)


app/serializers/service_provider_serializer.rb
```diff
@@ -0,0 +1,38 @@
...
+    :agency,
+    :friendly_name,
+    :metadata_url,
+    :acs_url,
+    :assertion_consumer_logout_service_url,
+    :cert,
+    :block_encryption,
+    #:return_to_sp_url,
+    #:logo,
+    #:attribute_bundle,
```
zachmargolis: can we add a `:redirect_url` -- I'd like to use that for the OpenID connect too

app/serializers/service_provider_serializer.rb
```diff
@@ -0,0 +1,38 @@
+class ServiceProviderSerializer < ActiveModel::Serializer
+  attributes(
```
zachmargolis: style-wise WDYT about dropping the parens?

config/routes.rb
```diff
@@ -19,5 +19,9 @@
     end
   end
 
+  namespace :api do
+    resources :service_providers, only: [:index]
```
zachmargolis: I really like the explicit path style we have in IDP repo, WDYT about keeping it for readability?

```
get '/api/service_providers' => 'api/service_providers#index'
```

app/serializers/service_provider_serializer.rb
```diff
@@ -0,0 +1,43 @@
...
+    :updated_at,
+    :signature
+  )
+
+  def agency
+    object.name
+  end
+
+  def friendly_name
+    object.description
```
zachmargolis: should we consider renaming some of these columns to match the keys that IDP expects? (minimize mental effort)

app/serializers/service_provider_serializer.rb
zachmargolis: (I mean comment it out as a placeholder similar to the other placeholders here)

app/serializers/service_provider_serializer.rb
pkarman: see https://github.com/18F/identity-private/issues/1311 -- I'm trying to keep the scope of this PR small. I agree though; should change db names.

app/serializers/service_provider_serializer.rb
pkarman: sure see 3d0814a

app/serializers/service_provider_serializer.rb
pkarman: sure see 3d0814a

app/serializers/service_provider_serializer.rb
pkarman: looks like ruby parser wants commas? I have added them back.

zachmargolis: the parser wants parens if they're on different lines, this would be acceptable, but it's neither here nor there:

```ruby
attributes :foo,
           :bar
```

app/serializers/service_provider_serializer.rb
zachmargolis: Sorry that was a forward-thinking wonder (ex when we make a migration to add the other columns we need, should we consider renaming these existing columns in the DB too?)

config/routes.rb
pkarman:
```diff
diff --git a/config/routes.rb b/config/routes.rb
index 8f281f1..104f400 100644
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -14,14 +14,16 @@ Rails.application.routes.draw do
     get 'timeout' => 'users/sessions#timeout'
   end
 
-  namespace :users do
-    resources :service_providers do
-    end
-  end
+  get '/users/service_providers' => 'users/service_providers#index'
+  post '/users/service_providers' => 'users/service_providers#create'
+  get '/users/service_providers/new' => 'users/service_providers#new'
+  get '/users/service_providers/:id/edit' => 'users/service_providers#edit'
+  get '/users/service_providers/:id' => 'users/service_providers#show'
+  patch '/users/service_providers/:id' => 'users/service_providers#update'
+  put '/users/service_providers/:id' => 'users/service_providers#update'
+  delete '/users/service_providers/:id' => 'users/service_providers#destroy'
 
-  namespace :api do
-    resources :service_providers, only: [:index]
-  end
+  get '/api/service_providers' => 'api/service_providers#index'
 
   root to: 'home#index'
 end
```

kinda like the terse rails magic, but respect to the explicit style. what do you think?

zachmargolis: I'm really enjoying how explicit we are with the paths (makes searching way easier)

But if the project already only does terse rails style, I'm fine with that. Do we want to enforce any consistency between idp and dashboard?

FWIW I prefer the explicit style, but this is not a big deal to me right now, I'd rather focus on shipping the  functionality.

app/serializers/service_provider_serializer.rb
amoose: Yep, Zach, Tom is working on that in #1311 👍

PR #46, opened 2017-01-13
Update attributes

blacktm:
This PR addresses https://github.com/18F/identity-private/issues/1311. Namely, it adds missing [attributes](https://pages.18f.gov/identity-dev-docs/#2-configuring-logingov-to-accept-your-sp) required by the IdP for integration, renames or modifies others for consistency, and fixes UI issues. This is all in service of having a tool for agencies to register their _test_ SPs, and have that data be exposed to our IdP for integration testing only.

The only remaining attribute to implement is the `attribute_bundle` array, but we'll do that in another PR.

### DIFF
--- a/app/serializers/service_provider_serializer.rb
+++ b/app/serializers/service_provider_serializer.rb
@@ -1,28 +1,19 @@
 class ServiceProviderSerializer < ActiveModel::Serializer
   attributes(
+    :friendly_name,
     :issuer,
     :agency,
-    :friendly_name,
-    :metadata_url,
     :acs_url,
     :assertion_consumer_logout_service_url,
-    :cert,
+    :sp_initiated_login_url,
     :block_encryption,
-    #:redirect_url,
-    #:return_to_sp_url,
+    :cert,
+    :return_to_sp_url,
     #:logo,
     #:attribute_bundle,
     :updated_at,
     :signature
   )
-
-  def agency
-    object.name
-  end
-
-  def friendly_name
-    object.description
-  end
 
   def cert
     object.saml_client_cert
--- a/app/views/users/service_providers/show.html.slim
+++ b/app/views/users/service_providers/show.html.slim
@@ -1,4 +1,4 @@
-h1 = t('headings.service_providers.show', name: service_provider.name)
+h1 = t('headings.service_providers.show', friendly_name: service_provider.friendly_name)
 
 .wrapper.actions
   = button_to t('forms.buttons.edit_service_provider'), edit_users_service_provider_url(service_provider),
@@ -8,17 +8,20 @@ h1 = t('headings.service_providers.show', name: service_provider.name)
 
 table.horizontal-headers
   tr
-    th Issuer
-    td.issuer = service_provider.issuer
-  tr
-    th Name
-    td.name = service_provider.name
+    th Friendly Name
+    td.friendly_name = service_provider.friendly_name
   tr
     th Description
     td.description = service_provider.description
   tr
-    th Metadata URL
-    td.metadata_url = service_provider.metadata_url
+    th Issuer
+    td.issuer = service_provider.issuer
+  tr
+    th Agency
+    td.agency = service_provider.agency
+  tr
+    th Logo
+    td.logo = ''
   tr
     th Assertion Consumer Service URL
     td.acs_url = service_provider.acs_url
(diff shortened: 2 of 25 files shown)


pkarman:
FWIW, the reason the issuer is non-editable and automatically a unique string is to guarantee that it is unique at the IdP as well. By putting the burden of unique-ness on the user, we require each user to come up with some unique value. Unless we provide hints or some conventions, that could be a problem, since user's cannot see one another's `ServiceProvider` entries.

amoose:
That's a good point, Peter. We'll be sure to consider this in the design, but we're going with a free text-field in order to permit the most flexibility for the user, right now, while still using proper data constraints. Also, the full list of SPs will be accessible. 

I like the idea of enforcing our own schema, prefixed with an agency name, or some other client tag. Let's continue this discussion here: https://github.com/18F/identity-private/issues/588

db/migrate/20170112194141_change_name_to_friendly_name.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeNameToFriendlyName < ActiveRecord::Migration
+  def change
+    rename_column :service_providers, :name, :friendly_name
```
zachmargolis: How are we planning on running this migration? Are we going to deploy code that aliases the column name first? +1/-1 compatibility? Or are we just going to take downtime while we migrate the DB?

pkarman: given that no existing dashboard instance should have any users or service provider records in it, I think it is safe to say we just deploy and suffer the downtime blip if it happens.

app/serializers/service_provider_serializer.rb
```diff
@@ -17,7 +17,7 @@ class ServiceProviderSerializer < ActiveModel::Serializer
   )
 
   def agency
-    object.name
+    object.friendly_name
```
pkarman: shouldn't this be `agency` instead -- in which case, no explicit method definition required for `agency` or `friendly_name` since they are now column names.

blacktm:
@pkarman Well, the agency need I'm attemping to address in the short term is to rapidly register/modify/delete SPs for integration testing _only_ (production SPs will be committed to the IdP repo, at least for now) with as little manual intervention from our team as possible. Tens, and soon hundreds, of CRUD operations will be flooding in, and we don't want to be a bottleneck. The dashboard will still be protected via basic auth, so the general public won't be able to create SPs. We could whitelist individual agency users, but that doesn't seem necessary to me. If you think there are risks to this approach, please let me know.

app/serializers/service_provider_serializer.rb
blacktm: Oh, shoot. Should be `agency`, so yes, can be removed. And same with `def friendly_name`, can also be removed — `description` is only used in the dashboard for agencies to add context to their SP entry. Thanks!

pkarman:
@blacktm understood, and sounds reasonable.

app/views/users/service_providers/show.html.slim
```diff
@@ -8,29 +8,41 @@ h1 = t('headings.service_providers.show', name: service_provider.name)
...
+    td.friendly_name = service_provider.friendly_name
   tr
     th Description
     td.description = service_provider.description
   tr
-    th Metadata URL
-    td.metadata_url = service_provider.metadata_url
+    th Issuer
+    td.issuer = service_provider.issuer
+  tr
```
pkarman: Might be worth re-ordering Agency and Friendly Name. The Agency is the default value on the IdP, which Friendly Name overrides just in UI. Putting Agency first would emphasize that, IMO.

Also, IdP treats Friendly Name as optional but not Agency. Should we put a not-null constraint on Agency?

amoose: +1 non-null constraint on Agency

amoose:
Yes, Peter, the approval mechanism will need to be fully vetted and approved before we enable this type of functionality in true-production. 👍

app/views/users/service_providers/show.html.slim
blacktm: Hmm, might be worth revisiting the list of SP attributes in our [dev docs](https://pages.18f.gov/identity-dev-docs/#2-configuring-logingov-to-accept-your-sp), see if we're presenting them accurately. Didn't know that about `agency` and `friendly_name` — here we have them as two totally different concepts.

And yes, there should probably be non-null constraint on `agency`, and actually we probably missed a few others that are required as well (see dev docs).

blacktm: Also regarding order of attributes, I went back and forth on that. I think agencies would appreciate consistency there (in the dashboard, dev docs, elsewhere), and it'll be convenient for us as well. Would love ideas on what people think the order should be.

pkarman: I hadn't seen those dev docs before -- helpful to know.

If we are going to rely on the `agency` being a string for grouping service providers together, then perhaps we need to create a model for that on the dashboard side for ensuring consistency in naming, etc. A single typo would break the grouping.

Also, an `Agency` model could function much like a traditional Organization and allow us to group `User` records together for authorization purposes. I.e. UserA and UserB are in the same `Agency` and so may edit each other's records.

pkarman: Re: ordering of attributes, if we are not going to alphabetize, then I would suggest going in an order similar to how personal contact info forms typically flow: name, address from specific to general, etc.

pkarman: I think we should merge this as-is and then create a new PR to refactor Agency into a model.

blacktm: Yes, great ideas re: Agency. Completely agree. For reordering, yes, I think if it follows a common sense pattern, then agencies will appreciate it.

pkarman:
@blacktm please squash into a single commit and I can merge.

PR #48, opened 2017-01-17
Add ServiceProvider.attribute_bundle

pkarman:
**Why**: attribute_bundle determines which asserted attributes are returned in an Authentication request


pkarman:
let's leave these commits un-squashed, as the travis change should be explicit in commit history.

PR #49, opened 2017-01-19
Remove unused gems

pkarman:
**Why**: Hygiene is a virtue


Gemfile
```diff
@@ -4,23 +4,14 @@ ruby '2.3.3'
 
 gem 'actionmailer-text'
 gem 'active_model_serializers'
-gem 'bourbon', '5.0.0.beta.5'
-gem 'coffee-rails', '~> 4.1.0'
```
zachmargolis: 🔥 🔥 🔥 🔥 

I believe at one point our zxcvbn code relied on that? But if the tests pass.... 💯

Gemfile
```diff
@@ -49,9 +40,7 @@ group :development do
 end
 
 group :development, :test do
-  gem 'awesome_print'
   gem 'bullet'
-  gem 'bundler-audit', '>= 0.5.0', require: false
```
zachmargolis: I believe this is the gem that @monfresh uses to upgrade our gems? So it's not used in the app, but it's used on the app?

Gemfile
zachmargolis: Oh wait this is `-dashboard` not `-idp` -- nevermind!

Gemfile
monfresh: No, you're thinking of `bummr`. `bundler-audit` is used to find CVEs and other Gemfile offenses. It's similar to Gemnasium, but it's already enabled in this repo's [Code Climate config](https://github.com/18F/identity-dashboard/blob/master/.codeclimate.yml#L4-L5), so unless devs need it to run locally, it's fine to remove.

PR #51, opened 2017-01-20
Add API endpoint to publish Service Providers

pkarman:
**Why**: Allow self-service one click to publish active SPs to IdP.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,3 +1,4 @@
 # env vars for CI only
+IDP_SP_URL=http://idp.example.com/api/service_provider
 IDP_SSO_URL=http://idp.example.com/saml/auth
 IDP_SLO_URL=http://idp.example.com/saml/logout
--- a/spec/services/service_provider_updater.rb
+++ b/spec/services/service_provider_updater.rb
@@ -1,0 +1,9 @@
+require 'rails_helper'
+
+describe ServiceProviderUpdater do
+  describe '#ping' do
+    it 'returns true for success' do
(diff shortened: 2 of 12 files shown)


.env.test
```diff
@@ -1,3 +1,4 @@
 # env vars for CI only
+IDP_URL=http://idp.example.com
 IDP_SSO_URL=http://idp.example.com/saml/auth
 IDP_SLO_URL=http://idp.example.com/saml/logout
```
amoose: I don't like mixing implicit and explicit ENV vars like this. I think either use a root host env var or just call the new var IDP_SP_ENDPOINT

pkarman: 👍  changed and pushed.

pkarman:
thanks @amoose -- comments addressed. I think the test coverage report might be misleading; there is a specific unit test in `spec/services/service_provider_updater.rb`

spec/services/service_provider_updater.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+describe ServiceProviderUpdater do
+  describe '#ping' do
+    it 'returns true for success' do
+      expect(subject.ping).to eq true
+    end
+  end
+end
```
monfresh: This filename is missing the `_spec` at the end, which is probably why CC thinks it's not tested.

pkarman: ah! good catch @monfresh -- I often make that mistake. Fixed in #54

PR #56, opened 2017-01-30
Re-enables signed requests with embedded signatures

amoose:
### Why
Logout requests are not being signed since embed_sign is set to false

### How
Enables embed_sign

### DIFF
--- a/lib/saml_config.rb
+++ b/lib/saml_config.rb
@@ -32,7 +32,7 @@ module Saml
         security: {
           authn_requests_signed: true,
           logout_requests_signed: true,
-          embed_sign: false,
+          embed_sign: true,
           digest_method: 'http://www.w3.org/2001/04/xmlenc#sha256',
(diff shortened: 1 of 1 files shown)


jessieay:
@amoose what do we need to do in order to get this merged?

pkarman:
I'm not sure we need this anymore if signatures are now supported in the URL on the IdP side.

We toggled `embed_sign` to `false` in 73b8d4fd to work around a URL length restriction in nginx when we were deployed to cloud.gov. The signature is still present in the request, it's just in a separate parameter, not embedded in the encoded SAML blob.

See https://github.com/18F/identity-idp/commit/b16237897116717928d19c99b6c826449f73a455

I'd like to close if that's ok with @amoose

jessieay:
Hi @amoose  -- per Peter's comment above, you OK with us closing this?

PR #57, opened 2017-02-02
Include all ServiceProviders in feed

pkarman:
**Why**: Active flag filter prevents IdP from removing de-activated
SPs from the IdP cache.

**How**: Let the IdP do its own filtering, by
including the active flag as an attribute in feed.


pkarman:
@amoose feel free to merge once you've had a look.

PR #58, opened 2017-02-04
Fix broken SP routes

blacktm:
**Why**: Since we're iterating on issuer formats (currently favoring an
URN), remove "issuer" as the ID in routing to avoid illegal URL
characters and broken routes.


amoose:
🤘 📈

PR #59, opened 2017-02-05
Show all SPs to admins

blacktm:
**Why**: As an admin, I want to see all registered agency SPs,
including the ones I’ve created, so that I can easily view and manage
them on behalf of the agency.


pkarman:
looks like a rebase and a squash is in order, then should be good to go, @blacktm

PR #62, opened 2017-02-08
Adds support for bastion host

amoose:
Why
To permit developers to deploy via capistrano

How
Enable SSH proxy (bastion) host

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -1,3 +1,4 @@
+require 'net/ssh/proxy/command'
 #################
 # GLOBAL CONFIG
 #################
@@ -18,7 +19,19 @@ set :passenger_restart_runner, :sequence
 set :rails_env, :production
 set :repo_url, 'https://github.com/18F/identity-dashboard.git'
 set :ssh_options, forward_agent: false, user: 'ubuntu'
+set :bastion_user, 'ubuntu'
 set :tmp_dir, '/tmp'
+
+set :ssh_options do
+  ssh_command = "ssh -A #{fetch(:bastion_user)}@#{fetch(:bastion_host)} -W %h:%p"
+    {
+    proxy: Net::SSH::Proxy::Command.new(ssh_command),
+    user: 'ubuntu',
+  }
+end
+
+server 'apps_host', roles: %w(web app db)
+
 
 #########
 # TASKS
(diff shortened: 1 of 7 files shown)


config/deploy.rb
```diff
@@ -18,8 +19,20 @@
 set :rails_env, :production
 set :repo_url, 'https://github.com/18F/identity-dashboard.git'
 set :ssh_options, forward_agent: false, user: 'ubuntu'
+set :bastion_user, 'ubuntu'
 set :tmp_dir, '/tmp'
 
+set :ssh_options do
+  ssh_command = "ssh -A #{fetch(:bastion_user)}@#{fetch(:bastion_host)} -W %h:%p"
+    {
```
zachmargolis: the linters are grumpy here about indendation

config/deploy.rb
```diff
@@ -18,8 +19,20 @@
...
+    proxy: Net::SSH::Proxy::Command.new(ssh_command),
+    user: 'ubuntu',
```
zachmargolis: should we reuse `fetch(:bastion_user)` here?

config/deploy.rb
```diff
@@ -18,8 +19,20 @@
 set :rails_env, :production
 set :repo_url, 'https://github.com/18F/identity-dashboard.git'
 set :ssh_options, forward_agent: false, user: 'ubuntu'
+set :bastion_user, 'ubuntu'
 set :tmp_dir, '/tmp'
 
+set :ssh_options do
+  ssh_command = "ssh -A #{fetch(:bastion_user)}@#{fetch(:bastion_host)} -W %h:%p"
```
zachmargolis: is there a way to override the bastion user on this line? I thought we were supposed to log in to the bastion as ourselves

config/deploy.rb
amoose: no, this is not the same user. jumphost <-> apps_host is as `ubuntu`

zachmargolis: got it, thanks!

config/deploy.rb
amoose: I will default it to ubuntu, but yes, I'll make it flexible for higher envs that require we auth via PIV 👍

amoose:
Tested in `qa` 🏁 👍

PR #70, opened 2017-02-27
Use Figaro to manage environment variables

jessieay:
**Why**: This is how other identity apps manage environment variables.


jessieay:
cc @amoose @nickbristow @blacktm

app/views/application/_analytics.html.erb
```diff
@@ -1,7 +0,0 @@
-<% if ENV["SEGMENT_KEY"] %>
```
nickbristow: No longer need analytics?

spec/rails_helper.rb
```diff
@@ -1,9 +1,8 @@
 ENV['RACK_ENV'] = 'test'
 
 require 'spec_helper'
-
 require File.expand_path('../../config/environment', __FILE__)
-abort('DATABASE_URL environment variable is set') if ENV['DATABASE_URL']
+# abort('DATABASE_URL environment variable is set') if Figaro.env.database_url
```
nickbristow: This wasn't commented out and now it is? Not sure what this was for.

app/views/application/_analytics.html.erb
jessieay: this was never used. I know that because 18F does not have an ATO for Segment :)

spec/rails_helper.rb
jessieay: good catch! commented this when debugging an issue with Figaro not being loaded but should have un-commented again. Did that at 2a211e3

app/views/application/_analytics.html.erb
nickbristow: 👍

zachmargolis:
Should we add an initializer with `Figaro.require_keys(...` to make sure we have all the env vars set that we want?

config/environments/production.rb
```diff
@@ -2,23 +2,23 @@
 Rails.application.configure do
   config.cache_classes = true
   config.eager_load = true
-  config.consider_all_requests_local       = false
+  config.consider_all_requests_local = false
   config.action_controller.perform_caching = true
-  config.serve_static_files = ENV['RAILS_SERVE_STATIC_FILES'].present?
+  config.serve_static_files = Figaro.env.serve_static_files == true
```
pkarman: Figaro values are always String objects afaik, so the `true` may need to be `'true'`.

config/puma.rb
```diff
@@ -12,15 +12,15 @@
 # Starting with a low number of workers and threads provides adequate
 # performance for most applications, even under load, while maintaining a low
 # risk of overusing memory.
-workers Integer(ENV.fetch('WEB_CONCURRENCY', 2))
-threads_count = Integer(ENV.fetch('MAX_THREADS', 2))
+workers Integer(Figaro.env.web_concurrency || 2)
```
pkarman: have you tested this works? I assumed that Puma was not reading from Figaro (i.e. the Rails app loads Figaro *after* puma has been initialized).

config/secrets.yml
```diff
@@ -1,7 +1,7 @@
 default: &default
   secret_key_base: MGhvYnpIWnJaRHhLaDJrRkpOd1JqWFhHUHMuRFpVeGRhRE5GVHY1NDdBQkwxM0NkWGdnY1NnZmhNMWhCRzBBRG80U2YuTG5hMURrb3NkZ2VURkJOelZkZmhFMUM0TWJwdGFwcFJhM21hRHdQVTFoRUpWZXZCbmp3YnROZkZ0a2MK
-  http_auth_username: <%= ENV["SP_NAME"] %>
-  http_auth_password: <%= ENV["SP_PASS"] %>
+  http_auth_username: <%= Figaro.env.sp_name %>
```
pkarman: I don't know if http auth config is necessary anymore now that dashboard is deployed behind nginx like IdP is. I would suggest we drop all the http auth config, but that can be a subsequent PR when `app/controllers/application_controller.rb` is also updated.

spec/rails_helper.rb
```diff
@@ -1,9 +1,8 @@
 ENV['RACK_ENV'] = 'test'
 
 require 'spec_helper'
-
 require File.expand_path('../../config/environment', __FILE__)
-abort('DATABASE_URL environment variable is set') if ENV['DATABASE_URL']
+abort('DATABASE_URL environment variable is set') if Figaro.env.database_url
```
pkarman: tbh I'm not sure why this was ever here. Smells like it derived from a template or cargo culting.

config/environments/production.rb
jessieay: good catch! updated at b0eb9a0

config/puma.rb
jessieay: silly me, I thought that if the tests passed that meant the app was running. I was wrong! :) 

Any suggestions on what to do here instead? Stick with `ENV`? Load `Figaro` ?

spec/rails_helper.rb
jessieay: yeah agreed. removed at d58fda2

config/puma.rb
monfresh: Are we even using Puma in production? I thought all our Rails apps were configured to use Passenger? https://github.com/18F/identity-dashboard/blob/7ec7dad09650c9b6c08ce81c7e216ffa5187acaf/config/deploy.rb#L16-L18

jessieay: hmmm interesting question. Looks like this was added back when this project was first created / deployed with Cloud.gov. Perhaps I will leave as `ENV` and follow up with a PR removing refs to Puma?

pkarman:
puma is used locally via Profile. Could switch to rackup like idp does.

config/secrets.yml
jessieay: will do! adding notes for what to do next now

config/puma.rb
jessieay: reverted at 685bf1c98ef0e32ce3ba490f7cc1a6e4a1fbe925 and will remove separately

jessieay:
@zachmargolis added required keys at  1308efe

config/newrelic.yml
```diff
@@ -10,7 +10,7 @@ common: &default_settings
     capture_source: true
     enabled: true
     ignore_errors: "ActionController::RoutingError,Sinatra::NotFound"
-  license_key: "<%= ENV["NEW_RELIC_LICENSE_KEY"] %>"
+  license_key: "<%= Figaro.env.new_relice_license_key %>"
```
zachmargolis: typo: `new_relice` sounds like a kind of licorice to me

jessieay:
@pkarman I like that idea, will do it in followup PR

config/newrelic.yml
jessieay: fixed - oy!

jessieay:
@zachmargolis yeah, but my guess is that right now this will be totally broken until we make the required updates to the devops repo...which is super confusing.

config/secrets.yml
jessieay: PR for this here https://github.com/18F/identity-dashboard/pull/74

monfresh:
Note that Figaro doesn't require the keys to be lowercase. You can use uppercase if you want. It's only a matter of preference and readability. You can do `Figaro.UPPERCASE_KEY` if you want.

jessieay:
@monfresh interesting, did not know that! Figured it had to be lowercase since that is what it is in all examples in the Figaro README and how we use Figaro in the IDP app.

Do you think I should update this PR to use uppercased env vars instead? It is def more standard to have ENV vars be uppercase, but I don't want to create confusion by having the dashboard and the idp app use Figaro differently, since the whole point if introducing Figaro here is to increase consistency...

monfresh:
If we want consistency, I would stick with lowercase.

zachmargolis:
I feel like, in terms of the smallest change we can ship, leaving the casing of the env vars alone is a good first step. Then we can deploy this, make sure it works, and then work to change the keys from there. We could say all new keys should be lowercase, but there are already uppercase ones that we use and have to coordinate changing in devops

jessieay:
@zachmargolis I like that suggestion. Agree w @monfresh that we eventually want consistency, but getting something that works is a good first step.

I am going to pair with @mugizico today on updating the env vars in devops land. I will work to get all env vars updated so we have upcase-d and downcase-d versions available. 

In the meantime, will update this PR to use uppercase env vars so that it can be safely merged and deployed without devops changes.

jessieay:
UPCASE-d at 60e4a60838865894b41934f4e0755d7c2ad8f79c

config/deploy.rb
```diff
@@ -4,7 +4,7 @@
 #################
 set :application, 'dashboard'
 # set branch based on env var or ask with the default set to the current local branch
-set :branch, ENV['branch'] || ENV['BRANCH'] || ask(:branch, `git branch`.match(/\* (\S+)\s/m)[1])
+set :branch, Figaro.env.BRANCH || ask(:branch, `git branch`.match(/\* (\S+)\s/m)[1])
```
zachmargolis: I'm not sure that figaro is loaded in the deploy tasks like this?

jessieay: good call -- should we just remove env var refs in here?

zachmargolis: These are for being able to override stuff, ex `BRANCH=mycustombranch cap staging deploy` to do a branch deploy, so I'd leave them

jessieay: 🆒  env var-ified at ef0854f

PR #71, opened 2017-02-28
Clean up rubocop violations

jessieay:
**Why**: Makes it easier to fix new style violations when existing
codebase does not have violations.


lib/saml_config.rb
```diff
@@ -2,12 +2,10 @@
...
   class Config
-    def initialize
-      @@settings ||= build_settings.freeze
-    end
+    attr_reader :settings
 
-    def settings
-      @@settings
+    def initialize
+      @settings ||= build_settings.freeze
```
pkarman: you've changed this from a class variable to an instance variable, which means that the config will be re-read from disk on every instantiation. That's what the class variable was trying to avoid.

We've solve this rubocop issue with `cattr_accessor` in the idp app. Maybe use that as a guide?

config/environments/development.rb
```diff
@@ -19,6 +19,9 @@
   config.action_mailer.delivery_method = :smtp
   config.action_mailer.raise_delivery_errors = true
   config.action_mailer.default_url_options = { host: 'localhost:3001' }
-  config.action_mailer.smtp_settings = { address: ENV['SMTP_HOST'] || 'localhost', port: ENV['SMTP_PORT'] || 2025 }
+  config.action_mailer.smtp_settings = {
+    address: ENV['SMTP_HOST'] || 'localhost',
+    port: ENV['SMTP_PORT'] || 2025
```
zachmargolis: Should we port the "require trailing comma" style from IDP?

lib/saml_config.rb
jessieay: ah gotcha. Rubocop was complaining about usage of a class var at all. Perhaps I should disable that cop?

config/environments/development.rb
jessieay: I'd love to do that! But perhaps in a separate PR so we can keep the separate threads separate (1 PR for updating styles for current rubocop config, one for updating rubocop config rules)

zachmargolis: Sounds good!

lib/saml_config.rb
pkarman: it's not disabled in idp (in .rubocop.yml) and no complaint about `cattr_accessor` there.

jessieay: getting this error when I change back to class var

`Replace class var @@settings with a class instance var.`

jessieay: are we sure that there's a difference in terms of re-reading from disk? I have been googling around it and seems like the only difference between instance vars and class vars is scope  availability http://stackoverflow.com/questions/15773552/ruby-class-instance-variable-vs-class-variable

zachmargolis: I think there are some wires crossed? @jessieay that blog post is talk about instance vars on a class object, I think @pkarman is talking about instance variables on an instance of a class.

We instantiate `Saml::Config` objects like this in at least 4 spots, two are inside of controllers, so they are per-request.

```
identity-dashboard:master> git grep Saml::Config.new -- {app,config} | wc -l
       4
```

So I think what @pkarman is suggesting is

```ruby
module Saml
  class Config
    cattr_accessor :settings do
      build_settings # promote this to a class method
    end

    def initialize
      @settings = self.class.settings.dup.freeze
    end
  end
end
```

monfresh: @jessieay You can test this by adding the line `puts 'build_settings called'` to the top of the `build_settings` method, and then running `rake`. You will notice that with the class variable version of the code, "build_settings called" only gets output once. With an instance variable, it gets called many times.

jessieay: @zachmargolis gotcha! updated and force pushed :)

monfresh: Personally, I would use the good old constant:

```ruby
require 'hashie/mash'

module Saml
  class Config
    # rubocop:disable AbcSize, MethodLength
    def self.build_settings
      puts 'build_settings called'
      config_file = "#{Rails.root}/config/saml.yml"
      saml_env_config = Hashie::Mash.new(YAML.load_file(config_file).fetch(Rails.env, {}))
      Hashie::Mash.new(
        issuer: ENV.fetch('SP_ISSUER', saml_env_config[:sp_issuer]),
        idp_sso_target_url: ENV.fetch('IDP_SSO_URL', saml_env_config[:sso_url]),
        idp_slo_target_url: ENV.fetch('IDP_SLO_URL', saml_env_config[:slo_url]),
        idp_cert_fingerprint: ENV.fetch('IDP_FINGERPRINT', saml_env_config[:idp_fingerprint]),
        name_identifier_format: ENV.fetch('IDP_NAME_ID_FORMAT', saml_env_config[:name_id_format]),
        certificate: ENV.fetch('SP_CERTIFICATE', saml_env_config[:sp_certificate]),
        private_key: OpenSSL::PKey::RSA.new(
          ENV.fetch('SP_PRIVATE_KEY', saml_env_config[:sp_private_key]),
          ENV.fetch('SP_PRIVATE_KEY_PASSWORD', saml_env_config[:sp_private_key_password])
        ).to_s,
        allowed_clock_drift: 5.minutes,
        authn_context: 'http://idmanagement.gov/ns/assurance/loa/1',
        single_signon_service_binding: 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST',
        double_quote_xml_attribute_values: true,
        security: {
          authn_requests_signed: true,
          logout_requests_signed: true,
          embed_sign: false,
          digest_method: 'http://www.w3.org/2001/04/xmlenc#sha256',
          signature_method: 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256'
        }
      )
    end
    # rubocop:enable AbcSize, MethodLength

    BUILD_SETTINGS = build_settings.freeze

    def logout_url
      settings[:idp_slo_target_url] || settings[:idp_sso_target_url].gsub(/auth$/, 'logout')
    end
  end
end
```

And then replace calls to `Saml::Config.new.settings` with `Saml::Config::SETTINGS`.

This has the advantage of being cleaner and only calling "build_settings` once, as opposed to the `cattr_accessor` solution, which calls it twice and is more verbose. Also, `cattr_accessor` is a Rails thing, not a Ruby thing.

jessieay: great suggestion! I am used to seeing constants used for this type of thing, so I think this direction sounds logical. Updated at 8489c34

PR #74, opened 2017-03-01
Remove HTTP auth config

jessieay:
**Why**: No longer needed since app is deployed behind nginx. See
https://github.com/18F/identity-dashboard/pull/70#discussion_r103475753

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -1,7 +1,5 @@
 default: &default
   secret_key_base: MGhvYnpIWnJaRHhLaDJrRkpOd1JqWFhHUHMuRFpVeGRhRE5GVHY1NDdBQkwxM0NkWGdnY1NnZmhNMWhCRzBBRG80U2YuTG5hMURrb3NkZ2VURkJOelZkZmhFMUM0TWJwdGFwcFJhM21hRHdQVTFoRUpWZXZCbmp3YnROZkZ0a2MK
-  http_auth_username: <%= ENV["SP_NAME"] %>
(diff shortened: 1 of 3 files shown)


config/secrets.yml
```diff
@@ -1,7 +1,5 @@
 default: &default
   secret_key_base: MGhvYnpIWnJaRHhLaDJrRkpOd1JqWFhHUHMuRFpVeGRhRE5GVHY1NDdBQkwxM0NkWGdnY1NnZmhNMWhCRzBBRG80U2YuTG5hMURrb3NkZ2VURkJOelZkZmhFMUM0TWJwdGFwcFJhM21hRHdQVTFoRUpWZXZCbmp3YnROZkZ0a2MK
-  http_auth_username: <%= ENV["SP_NAME"] %>
```
zachmargolis: Removing it from here will break the dashboard deploy status checker, we use these secrets to load the status deploy JSON files in the status checker: https://github.com/18F/identity-dashboard/blob/master/app/services/deploy_status_checker.rb#L89

jessieay: ah ok, so these are being used to basic auth to the IDP endpoints? 

Perhaps we can save them as regular env vars / Figaro env vars rather than putting them in `secrets.yml` ?

zachmargolis: Yeah, I don't mind where they go or get loaded from, just need to make sure we don't remove them from the app 😀

jessieay: thoughts on 5b01376 ?

zachmargolis: 💯

PR #78, opened 2017-03-02
Move agency back to being a string

jessieay:
**Why**: SPs will now belong to an Organization, which will be how we
group them. This model was added for grouping purposes. See convo at
https://github.com/18F/identity-dashboard/pull/46#discussion_r96485084

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -14,6 +14,7 @@ engines:
       - javascript
     exclude_paths:
     - 'spec/**/*'
+    - 'db/migrate/*'
   eslint:
     enabled: true
   fixme:
--- a/app/controllers/users/service_providers_controller.rb
+++ b/app/controllers/users/service_providers_controller.rb
@@ -90,7 +90,7 @@ module Users
       params.require(:service_provider).permit(
         :acs_url,
         :active,
-        :agency_id,
+        :agency,
         :approved,
         :assertion_consumer_logout_service_url,
         :block_encryption,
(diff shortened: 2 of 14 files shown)


pkarman:
I'm -1 on this, unless there's some business requirement I don't understand. I prefer FKs for consistent naming. The original change wasn't about grouping for authz purposes. It was about consistent naming.

jessieay:
@pkarman if we want consistency, we can use `agency_name` field from the organization, which is going to have 1:many relationship with service providers as of https://github.com/18F/identity-dashboard/pull/79 

I guess if we really care about consistency, we can have organizations belong to an agency. But I guess I am not clear on the business req for consistent agency names? Is that even a field we use? For example, for the micro-purchase service provider I set up the agency as `TTS Acquistion` but given the way that agency names work here, I would have needed to add a record to the seeds in order to get that in the DB: https://github.com/18F/identity-idp/blob/master/config/service_providers.yml#L81

Seems easier to just let it be a string?

nickbristow:
We may need @blacktm's input here.

pkarman:
Guess what I'm saying is, db normalization with FKs can serve multiple purposes. The authz structure (a User can belong to multiple Organizations) is orthogonal to the naming structure (a ServiceProvider has one Agency). The Agency table is, imo, like an enum that is stored in the db rather than code. That's all. Using the FK just enforces the data integrity. It doesn't imply any relationship between a User and an Agency or an Agency and an Organization.

jessieay:
@pkarman good chatting with you this morning. For others' edification:

The only reason we care about agency is that is shows up on a user's profile in the IDP re: which agencies they've logged in with. So we do kind of want some normalization for agencies. Otherwise, not very important.

Closing this

PR #79, opened 2017-03-02
Admin can add service provider to organization

jessieay:
* Via new and edit form

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -3,4 +3,8 @@ class Organization < ActiveRecord::Base
   validates :department_name, presence: true
   validates :team_name, presence: true
   validates :agency_name, uniqueness: { scope: [:department_name, :team_name] }
+
+  def to_s
+    team_name
+  end
 end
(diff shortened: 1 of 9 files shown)


jessieay:
cc @nickbristow

app/models/organization.rb
```diff
@@ -3,4 +3,8 @@ class Organization < ActiveRecord::Base
   validates :department_name, presence: true
   validates :team_name, presence: true
   validates :agency_name, uniqueness: { scope: [:department_name, :team_name] }
+
+  def to_s
+    "#{agency_name} | #{department_name} | #{team_name}"
```
nickbristow: This should be `department_name | agency_name | team name

jessieay: I am not 100% sure should have both `department` and `agency` -- maybe it should be an OR? 

see https://en.wikipedia.org/wiki/Government_agency#United_States

jessieay: Asked on Slack about this...OH WOW

jessieay: Conclusion from Slack: there is no consistency in this because government. We should combine the fields. Will do that separately.

PR #80, opened 2017-03-02
Make service provider routing more Railsy

jessieay:
**Why**:
* No reason to nest inside `user` since we know who the current user is
via Devise
* This makes the routes more consistent and clean, imho


config/routes.rb
```diff
@@ -14,18 +14,8 @@
...
-  get '/users/service_providers/new' => 'users/service_providers#new',
-      as: 'new_users_service_provider'
-  get '/users/service_providers/:id/edit' => 'users/service_providers#edit',
-      as: 'edit_users_service_provider'
-  get '/users/service_providers/:id' => 'users/service_providers#show', as: 'users_service_provider'
-  patch '/users/service_providers/:id' => 'users/service_providers#update'
-  put '/users/service_providers/:id' => 'users/service_providers#update'
-  delete '/users/service_providers/:id' => 'users/service_providers#destroy'
-
-  resources :organizations, only: [:new, :create, :show, :edit, :update, :destroy, :index]
```
zachmargolis: I really like the `only: []` syntax for highlighting which routes actually get used. WDYT of keeping it? It also makes `rake routes` cleaner because it doesn't show the ones we don't use

jessieay: @zachmargolis I like it too, but since we are using *all* routes it wasn't giving us much in terms of cleanliness. Should I change back anyway?

jessieay: Going to keep as-is for now. We can add back if you like! 🌈

zachmargolis: :+1:

PR #81, opened 2017-03-02
Add validation on service provider / agency relation

jessieay:
**Why**: Otherwise, there is no form validation but there is a postgres
error when submitting a form without agency selected


jessieay:
cc @nickbristow

PR #83, opened 2017-03-02
Add a read_timeout to DeployStatusChecker

zachmargolis:


zachmargolis:
Confirmed that this fixes the issue with the entire page timing out when viewing when downstream services are down

PR #88, opened 2017-03-03
Rename Organizations to UserGroups

jessieay:
**Why**: Organizations sounded too much like it was an official part of
government. This way, it's clear that we are just representing groups of
users to manage service providers. We can include agency / org /dept
info in the "description" field


nickbristow:
Looks good to me!

PR #90, opened 2017-03-03
Add "Manage user groups" link to nav for admin users

nickbristow:


jessieay:
I think you had a test for this in the other PR?

spec/features/user_groups_spec.rb
```diff
@@ -72,3 +72,25 @@
...
 end
+
+feature 'Nav links include link to User Groups' do
+  context 'user is admin' do
+    scenario 'admin should see manage user groups link' do
+      admin = create(:admin)
+      login_as(admin)
+      visit service_providers_path
+
+      expect(page).to have_content('Manage user groups')
```
jessieay: can we reference the i18n string here instead?

nickbristow: oh good point!

PR #91, opened 2017-03-06
Update omniauth, omniauth-saml, and hashie

nickbristow:
Hashie is surfacing an error 

This error is coming from an Omniauth depency. They are aware of the issue. 

Removing hashie from the gem file, updated gems

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,6 @@ gem 'dotenv-rails'
 gem 'enum_help'
 gem 'figaro'
 gem 'foreman', require: false
-gem 'hashie'
 gem 'httparty'
 gem 'jquery-rails'
 gem 'newrelic_rpm', '>= 3.9.8'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -192,9 +192,9 @@ GEM
     newrelic_rpm (3.18.1.330)
     nokogiri (1.6.8.1)
       mini_portile2 (~> 2.1.0)
-    omniauth (1.3.1)
-      hashie (>= 1.2, < 4)
-      rack (>= 1.0, < 3)
+    omniauth (1.6.1)
+      hashie (>= 3.4.6, < 3.6.0)
+      rack (>= 1.6.2, < 3)
     omniauth-saml (1.7.0)
       omniauth (~> 1.3)
       ruby-saml (~> 1.4)
@@ -403,7 +403,6 @@ DEPENDENCIES
   factory_girl_rails
   figaro
   foreman
-  hashie
   httparty
   i18n-tasks
   jquery-rails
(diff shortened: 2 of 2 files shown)


nickbristow:
More info. https://github.com/omniauth/omniauth/issues/872

monfresh:
To be more clear, including `hashie` in the Gemfile is redundant because it is already provided by the `omniauth` gem, which in turn is provided by the `omniauth-saml` gem. Removing `hashie` from the Gemfile does not remove `hashie` from this app because it is needed by `omniauth`.

The issue that @jessieay had reported has been fixed in the latest versions of omniauth and hashie. To fix the issue, we need to update omniauth to the latest version.

monfresh:
As for my comment about getting rid of hashie, I was not talking about the gem itself (I had not realized omniauth depends on it). I was talking about our use of `Hashie::Mash` in `lib/saml_config.rb`. As far as I know, this line in the devise config:
```ruby
  config.omniauth :saml, 
```
only expects a regular Hash after the comma.

`Hashie::Mash` is slow and is rarely beneficial compared to a plain Hash or a PORO (plain old Ruby object). Read this article for a compelling argument against the use of `Hashie::Mash`: http://www.schneems.com/2014/12/15/hashie-considered-harmful.html

pkarman:
FWIW, I chose to use `Hashie::Mash` for the SAML config because (a) Omniauth was already using it, and (b) ruby-saml (which Omniauth delegates to) seemed to require methods, not just a plain Hash.

The performance isn't really an issue here IMO because it gets used once, at app start time.

No objection to replacing it; just wanted to provide some context.

nickbristow:
@pkarman @monfresh The branch name doesn't make sense now, but I updated omniauth, omniauth-saml, and hashie to their latest versions. You were right, that stopped the error from showing up on rspec.

Gemfile
```diff
@@ -10,12 +10,13 @@ gem 'dotenv-rails'
 gem 'enum_help'
 gem 'figaro'
 gem 'foreman', require: false
-gem 'hashie'
+gem 'hashie', '~> 3.5', '>= 3.5.5'
```
jessieay: I think this can just be `gem 'hashie', '>= 3.5.5'` -- I am not sure what bundler does with two args, usually just see 1

nickbristow: yeah, i've never seen that either. Thats what you get when you copy from rubygems.org

jessieay: weird...let's stick with my suggestion?

nickbristow: So it looks like this means, 
highest version of 3.5 and higher than or equal to 3.5.5

I think its the same as `~> 3.5.5`

nickbristow: Removing version numbers based on @jessieay's advice

jessieay: CI is mad at you! :(

monfresh: We don't need `hashie` in the Gemfile at all. It's already included in the project by omniauth.

jessieay: truth!

PR #92, opened 2017-03-07
Add default value for `description` of empty string

jessieay:
* Enables devs with existing data to migrate


jessieay:
cc @nickbristow

PR #93, opened 2017-03-07
Update index page to show user group information

nickbristow:
Adds edit and delete links to index page
  remove show user_groups page
  Remove "_" from "Create new user_group"
  updated tests


spec/features/user_groups_spec.rb
```diff
@@ -22,7 +22,10 @@
     org = create(:user_group)
     login_as(admin)
 
-    visit edit_user_group_path(org)
+    visit user_groups_path
+    find("a[aria-label='#{t('links.aria.edit', name: org.name)}']").click
+
+    expect(current_path).to eq(edit_user_group_path(org))
```
jessieay: I don't know if you even need this expectation. The fact that the following steps work means that you are on the correct page (because there are name and description form fields). what do you think?

config/routes.rb
```diff
@@ -15,7 +15,7 @@
   end
 
   resources :service_providers
-  resources :user_groups
+  resources :user_groups, except: [:show]
```
jessieay: 👏

spec/features/user_groups_spec.rb
```diff
@@ -22,7 +22,10 @@
     org = create(:user_group)
     login_as(admin)
 
-    visit edit_user_group_path(org)
+    visit user_groups_path
+    find("a[aria-label='#{t('links.aria.edit', name: org.name)}']").click
```
jessieay: why not just do `click_on org.name`?

jessieay:
I think that in the controller we redirect to the show page after create, so I am guessing CI will fail at this point. Should we also remove the `show` action from the controller and remove the view template?

spec/features/user_groups_spec.rb
nickbristow: I think there are multiple user_groups on the page, i was thinking it just checks and make sure the link goes to the right user group

nickbristow:
@jessieay if it fails, we did good tests

spec/features/user_groups_spec.rb
nickbristow: I set it up with links that are just "Edit" and "Delete", when running tests, i found that there were multiple user_groups on the page.

PR #95, opened 2017-03-07
Add service providers to user group index page

nickbristow:


app/models/user_group.rb
```diff
@@ -3,6 +3,7 @@ class UserGroup < ActiveRecord::Base
 
   validates :description, presence: true
   validates :name, presence: true, uniqueness: true
+  has_many :service_providers
```
jessieay: can we add a quick shoulda spec for this in `spec/models/user_group_spec.rb` ?

PR #96, opened 2017-03-07
Only run simplecov in CI env

jessieay:
**Why**:
* When running `rpsec spec/ --profile`, i18n-tasks was slowest file
* Ran 3 times, each time file took ~2 seconds to run
* Upon investigation, this appears to relate to Simplecov
* See https://github.com/glebm/i18n-tasks/issues/221
* Also https://github.com/colszowka/simplecov/issues/404
* Since there is no proposed solution at the moment (slowness appears to
  be coming from `Coverage`, which is part of the Ruby stdlib), the
  short term solution is to only run Simplecov on Travis, where required
* New timing for i18n-tasks: under .2 seconds!!!

Before:

```
Top 10 slowest example groups:
  I18n
    1.8 seconds average (3.59 seconds / 2 examples) ./spec/i18n_spec.rb:4
  Users::ServiceProviders
    0.17479 seconds average (1.22 seconds / 7 examples) ./spec/requests/service_providers_spec.rb:3
```

After:

```
Top 10 slowest example groups:
  Users::ServiceProviders
    0.16671 seconds average (1.17 seconds / 7 examples) ./spec/requests/service_providers_spec.rb:3
  I18n
    0.14545 seconds average (0.29089 seconds / 2 examples) ./spec/i18n_spec.rb:4
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,10 @@ script:
   - make test
   - bundle exec codeclimate-test-reporter
 env:
-  - RACK_ENV=test
-  - RAILS_ENV=test
+  global:
+    - RACK_ENV=test
+    - RAILS_ENV=test
+    - COVERAGE=true
 addons:
   postgresql: 9.3
   code_climate:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,6 @@ script:
   - bundle exec codeclimate-test-reporter
 env:
   global:
-    - RACK_ENV=test
-    - RAILS_ENV=test
     - COVERAGE=true
 addons:
   postgresql: 9.3
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -16,7 +16,6 @@ RSpec.configure do |config|
   config.include Warden::Test::Helpers
 
   config.before(:suite) do
-    Rails.application.load_seed
     Warden.test_mode!
   end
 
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -13,8 +13,9 @@ script:
   - make test
   - bundle exec codeclimate-test-reporter
 env:
-  - RACK_ENV=test
-  - RAILS_ENV=test
+- RACK_ENV=test
+- RAILS_ENV=test
+- COVERAGE=true
```
jessieay: OMG we were running one build per env var before

> When you define multiple variables per line in the env array (matrix variables), one build is triggered per item.

https://docs.travis-ci.com/user/environment-variables/

monfresh: I don't think that was the case. That only applies for matrixed builds and when more than one ENV vars are defined *on the same line*. We were only running one version of Ruby, and we only had one ENV var per line. You can see that only one build was being run here: https://travis-ci.org/18F/identity-dashboard/builds

jessieay: Check out the two builds for this PR: https://travis-ci.org/18F/identity-dashboard/builds/208771094

and then after my PR https://travis-ci.org/18F/identity-dashboard/builds/208784528

jessieay: I am pretty confident this is the case because after this PR the two builds for `RACK_ENV` and `RAILS_ENV` failed because they did not have `COVERAGE` set

monfresh: You're right! I missed that. According to the docs, those ENV vars are already set by default: https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

jessieay: cool! I think we should keep the `global` namespace just so this problem doesn't get re-introduced in the future, but I removed those at bd3792b

PR #97, opened 2017-03-08
Remove gemset file

jessieay:
* We are using bundler, Gemset not needed. Haven't seen a gemset in the
wild in a few years :)

### DIFF
--- a/.ruby-gemset
+++ b/.ruby-gemset
@@ -1,1 +1,0 @@
-identity-dashboard


pkarman:
-1

the gemset file keeps all your bundler-maintained gems in an app-specific project directory, so that different versions of the same gem can co-exist on the same machine.

there's one for identity-idp too.

monfresh:
The ability to have different versions of the same gem on the same machine is what Bundler provides. A gemset is not required for that functionality. You can verify this by uninstalling RVM and using chruby and ruby-install instead (which is what the laptop script recommends now). `chruby` doesn't recognize the `.ruby-gemset` file, yet I am able to `bundle install` in different projects and have different gems for each project.

pkarman:
could we compromise? remove the `.ruby-gemset` file and add it to `.gitignore` so that those of us who use one can continue to.

monfresh:
Sounds good to me. I don't have a preference either way.

jessieay:
@pkarman I am not clear on what functionality you are getting from gemsets that you do not get from bundler? If there is something, I'd love to know since my current understanding is that [gemsets provide no value if you are using blinder](http://stackoverflow.com/a/4691611) ( As Moncef said, you can have different versions of the same gem on the same machine with bundler, no gemset required. 

Gemsets are also RVM specific, so would not be useful us weirdos who use Rbenv.

Finally, if you *really* want to use gemsets, can't you just add this file to your global `.gitignore` ?

pkarman:
I like the gemset approach because it namespaces my gem sets.
```
~/src/identity-dashboard $ bundle show devise
/Users/peterkarman/.rvm/gems/ruby-2.3.3@identity-idp/gems/devise-4.2.0
```

Now I know that instance of Devise is separate than the dashboard instance:
```
~/src/identity-dashboard $ bundle show devise
/Users/peterkarman/.rvm/gems/ruby-2.3.3@identity-dashboard/gems/devise-4.2.0
```

So if I hand-edit a gem in order to debug something, I know I'm affecting only the gem instance for only that app, not all the apps on my machine that happen to be using the same version of ruby.

This is a personal preference thing. Happy not to impose on others.

I see now that the idp version is, in fact, my own local copy, since `.ruby-gemset` is explicitly in the identity-idp `.gitignore` file (as I was suggesting we do here as well).

jessieay:
@pkarman the hand-editing thing is interesting! I can see how that gemsets would be helpful in that case. At the same time, editing gem source code locally is usually for temporary debugging so I don't think I am going to join the church of gemsets just yet :)

PR #99, opened 2017-03-08
Remove css lintrc file

jessieay:
**Why**: we do not use this.


jessieay:
haha! based on code climate error it looks like we ARE using this. closing :)

PR #100, opened 2017-03-08
Add multi-select input for adding users to user groups

jessieay:
* This is just using simple_form's basic multiselect for now
* Select and press shift to select another
* We can do something fancier later?

screen shot 2017-03-08 at 2 32 45 pm

### DIFF
--- a/app/views/user_groups/_form.html.slim
+++ b/app/views/user_groups/_form.html.slim
@@ -1,3 +1,3 @@
 = form.input :name
 = form.input :description
-= form.association :users, label_method: :to_s, collection: User.all
+= form.association :users, label_method: :to_s, collection: User.sorted
--- a/app/views/user_groups/_form.html.slim
+++ b/app/views/user_groups/_form.html.slim
@@ -1,2 +1,3 @@
 = form.input :name
 = form.input :description
+= form.association :users, label_method: :to_s, collection: User.all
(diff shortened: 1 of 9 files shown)


app/views/user_groups/_form.html.slim
```diff
@@ -1,2 +1,3 @@
 = form.input :name
 = form.input :description
+= form.association :users, label_method: :to_s, collection: User.all
```
jessieay: we should probably order these somehow. Since we do not have first name and last name for all users, perhaps by email address?

nickbristow: sounds good to me!

spec/features/user_groups_spec.rb
```diff
@@ -3,12 +3,14 @@
...
     admin = create(:admin)
-    login_as(admin)
+    user = create(:user)
 
+    login_as(admin)
     visit new_user_group_path
 
     fill_in 'Description', with: 'department name'
     fill_in 'Name', with: 'team name'
+    select user.email, from: 'Users'
```
nickbristow: should we check here for the user?

jessieay: what do you mean? like check that the user is added to the group? I figured we'd do that when we show the users on the user group index page....which I think you are working on?

nickbristow: Ah yes, not exactly what i was working on, but yes it can be checked then.

jessieay:
@nickbristow added users to user group index at 1866661

jessieay:
@nickbristow ready to re-review!

blacktm:
Looks great to me!

PR #101, opened 2017-03-08
Users can see and edit SP's associated with their user group

nickbristow:


app/policies/service_provider_policy.rb
```diff
@@ -44,7 +44,11 @@ def admin?
     current_user.admin?
   end
 
-  def owner_or_admin?
-    owner? || admin?
+  def member?
+    sp.user_group.present? && sp.user_group == current_user.user_group
```
jessieay: I think we want to let them edit ones that they created too, right? like what if I create a service provider but am somehow not in the user group I assign to it?

spec/features/service_providers_spec.rb
```diff
@@ -117,3 +117,40 @@
...
+    scenario 'service providers from a user group show on index' do
+      group = create(:user_group)
+      user1 = create(:user, user_group: group)
+      user2 = create(:user)
+      app = create(:service_provider, user_group: group, user: user2)
+      login_as(user1)
+
+      visit service_providers_path
+
+      expect(page).to have_content(app.friendly_name)
```
jessieay: should we also create a service provider for a different user group and ensure it is not showing on this page?

spec/features/service_providers_spec.rb
```diff
@@ -117,3 +117,40 @@
...
+    end
+
+    scenario 'user can edit a service provider from their user group', :js do
```
jessieay: should we add a scenario that a user can edit a service provider that they create?

why is js on here? better to turn off if we can for speed reasons ⏩

spec/features/service_providers_spec.rb
```diff
@@ -117,3 +117,40 @@
...
+      group = create(:user_group)
+      user1 = create(:user, user_group: group)
+      user2 = create(:user)
+      app = create(:service_provider, user_group: group, user: user2)
+      login_as(user1)
+
+      visit edit_service_provider_path(app)
+      fill_in 'Friendly name', with: 'change service_provider name'
```
jessieay: might be easier to name these strings something like `new_name` (variable) so we can just check for that variable in the expectations.

app/models/user.rb
```diff
@@ -47,6 +48,17 @@ def sync_with_auth_hash!(auth_hash)
     self
   end
 
+  def scoped_service_providers
```
jessieay: let's write a unit test for this!

app/models/user.rb
```diff
@@ -47,6 +48,17 @@ def sync_with_auth_hash!(auth_hash)
...
+    (
+      service_providers +
+      member_service_providers
+    ).uniq
+  end
+
+  def member_service_providers
```
jessieay: can this method be private?

nickbristow: yes. it can!

spec/features/service_providers_spec.rb
nickbristow: woops, turned it on for debugging

nickbristow: We are testing for editing a sp they created above

app/policies/service_provider_policy.rb
nickbristow: yes... :54 on the feature spec should be failing? But its not

PR #103, opened 2017-03-09
Fix slim-lint errors, run slim-lint in Travis

jessieay:
**Why**: This is what the IDP app does. We cannot run it via code
climate yet, see https://github.com/codeclimate/codeclimate/issues/418
Also adding a slim lint config file so we are following the same rules
as the IDP app


monfresh:
Looks like you haven't added the slim lint config yet. Travis is complaining about the 80-char line length.

jessieay:
@monfresh woops forgot to check it in. Did so at  876d5f8

PR #109, opened 2017-03-20
Show users who are already in a user group when editing

jessieay:
* There was a bug where we were showing only users without a user group
* This lead to an issue where users in a user group were not showing
  when editing that user group, leading to them being unassigned
* Now this is fixed and I added a regression test

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -7,7 +7,10 @@ class User < ActiveRecord::Base
   after_create :send_welcome
 
   scope :sorted, -> { order(email: :asc) }
-  scope :without_user_group, -> { where(user_group: nil) }
+
+  def self.without_other_user_group(user_group_id: nil)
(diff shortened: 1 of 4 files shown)


jessieay:
@zachmargolis I wasn't sure about that! Having both a feature spec and a unit test felt like double coverage to me, a bit, but I also don't want to under test. How do you think about feature spec coverage vs unit test coverage in cases like these?

zachmargolis:
My rule of thumb is I like to look for a test in the matching `_spec` file? So for a method defined in `app/models/user.rb` I would expect to find a test in `spec/models/user_spec.rb`.

I also don't mind testing something twice, because the feature specs catch different things?

jessieay:
@zachmargolis added unit tests at 651a3e5

I also expect this although since this method was so simple, I got it working by just making the feature spec pass...so I wasn't sure. Thanks for the nudge

PR #112, opened 2017-03-23
Show more environments

zachmargolis:
🎉 

We can disable the checks for things like SPs in production by removing the URLs and they'll appear grayed out, I'm not sure what apps exist in which environments.

identity_dashboard_

### DIFF
--- a/app/services/deploy_status_checker.rb
+++ b/app/services/deploy_status_checker.rb
@@ -5,6 +5,21 @@ class DeployStatusChecker
   Environment = Struct.new(:env, :statuses)
 
   DEPLOYS = [
+    Deploy.new('identity-idp', 'prod', 'https://idp.prod.login.gov'),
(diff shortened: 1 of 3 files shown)


app/services/deploy_status_checker.rb
```diff
@@ -5,6 +5,21 @@ class DeployStatusChecker
   Environment = Struct.new(:env, :statuses)
 
   DEPLOYS = [
+    Deploy.new('identity-idp', 'prod', 'https://idp.prod.login.gov'),
```
pkarman: will this be an alias to `secure.login.gov` at some point? or is that a different env?

zachmargolis: It should, but that name change hasn't been merged/deployed yet, so I figured it made sense to deploy the environment we currently have?

PR #113, opened 2017-03-24
Added confirm dialog box for deleting SP's

nickbristow:


jessieay:
is there a ticket for this? If not, let's make one. I think we should make sure there are GH tickets to track all of our work for the dashboard.

config/locales/en.yml
```diff
@@ -11,6 +11,7 @@ en:
       edit_service_provider: Edit
       trigger_idp_refresh: Publish service providers
     required_field: Indicates a required field.
+    confirm: 'Are you sure?'
```
jessieay: maybe "Are you sure you want to delete this service provider? This cannot be undone."

nickbristow:
@jessieay yep. https://github.com/18F/identity-private/issues/1707

jessieay:
Cool, I guess it didn't link back here (which is how I usually look for it) because you linked to the PR with the `/files` dir instead of the PR itself. Weird!

PR #114, opened 2017-03-24
Gem updates

jessieay:
via bummr


jessieay:
argh these updates are giving us new security issues -- any insight, @pkarman @monfresh ?

monfresh:
Huh. Very strange. Why would it report it here and not in the IdP repo? I bet most if not all of these are false positives. I looked at this one, for example: https://snyk.io/vuln/SNYK-RUBY-ACTIVERECORD-20270. That has been fixed in Rails 4.2.7.1 but Snyk seems to think it affects anything greater than or equal to 4.2.8 but less than 5.0.0.

monfresh:
I reported the bug to Snyk and they said they fixed it. Try force pushing and see if it clears it up.

jessieay:
@monfresh yay! sneak passed.

PR #115, opened 2017-03-28
Added user group select for all users

nickbristow:
Filters to users user group or nothing
  New SP's default to users user group
  Admin's can still select any user group

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,6 +12,14 @@ class User < ActiveRecord::Base
     where(user_group_id: [nil, user_group_id])
   end
 
+  def scoped_user_groups
+    if admin?
+      UserGroup.all
+    else
+      [user_group]
+    end
+  end
+
   def create_uuid
     self.uuid = SecureRandom.uuid unless uuid.present?
   end
--- a/app/views/service_providers/_form.html.slim
+++ b/app/views/service_providers/_form.html.slim
@@ -2,7 +2,12 @@
 
 - if current_user.admin?
   = form.input :approved, as: :radio_buttons
-  = form.association :user_group
+
+= form.association :user_group,
+  as: :collection_select,
+  collection: current_user.scoped_user_groups,
+  include_blank: '',
+  selected: selected_user_group
 
 = form.input :friendly_name
 = form.input :description
--- a/app/views/service_providers/edit.html.slim
+++ b/app/views/service_providers/edit.html.slim
@@ -1,7 +1,8 @@
 h2 = t('headings.service_providers.edit')
 
 = simple_form_for(service_provider, html: { autocomplete: 'off', role: 'form' }) do |form|
-  == render 'form', form: form
+  - selected_user_group = @service_provider.user_group_id
+  == render 'form', form: form, selected_user_group: selected_user_group
   .wrapper.actions
     .action
       = form.button :submit, 'Update'
--- a/app/views/service_providers/new.html.slim
+++ b/app/views/service_providers/new.html.slim
@@ -5,5 +5,6 @@ p = t('links.attr_description_html',
     'https://pages.18f.gov/identity-dev-docs'))
 
 = simple_form_for(@service_provider, html: { autocomplete: 'off', role: 'form' }) do |form|
-  == render 'form', form: form
+  - selected_user_group = current_user.user_group_id
+  == render 'form', form: form, selected_user_group: selected_user_group
   = form.button :submit, 'Create'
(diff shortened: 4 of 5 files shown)


app/views/service_providers/_form.html.slim
```diff
@@ -1,8 +1,21 @@
...
+  = form.association :user_group,
+    as: :collection_select,
+    collection: current_user.scoped_user_groups,
+    include_blank: ''
+-else
+  = form.association :user_group,
+    as: :collection_select,
+    collection: current_user.scoped_user_groups,
+    include_blank: '',
+    selected: current_user.user_group_id
```
jessieay: I think you can avoid this whole if / else by setting a default value of `current_user.user_group` like: http://stackoverflow.com/a/19029403

app/views/service_providers/_form.html.slim
```diff
@@ -1,8 +1,21 @@
+- edit_form ||= false
```
jessieay: let's use `content_for` instead http://guides.rubyonrails.org/layouts_and_rendering.html#using-the-content-for-method

jessieay: but if we don't need the if / else, per my other comment, then no need to use `content_for` !

spec/features/service_providers_spec.rb
```diff
@@ -96,3 +95,29 @@
     expect(page).to have_content('Success')
   end
 end
+
+feature 'User Group is set based on the users user group' do
```
jessieay: nice tests! I like to keep feature specs to one feature per file. Put these in their own file? or incorporate into another feature spec as sub-scenarios?

spec/features/service_providers_spec.rb
```diff
@@ -27,48 +26,74 @@
...
+
+      expect(page).to have_content('Success')
+      within('table.horizontal-headers') do
+        expect(page).to have_content('app description foobar')
+        expect(page).to have_content('change service_provider name')
+        expect(page).to have_content('last_name')
+      end
+    end
+
+    scenario 'service provider user group isnt overwritten by users user group on edit' do
```
jessieay: how about 

```
context 'service provider does not have a user group' do
  scenario 'user group defaults to nil' do
  end
end
```

spec/features/service_providers_spec.rb
```diff
@@ -27,48 +26,74 @@
...
-
-        expect(page).to have_content('Success')
-      end
+    scenario 'user group defaults to users user group' do
+      ug = create(:user_group)
+      user = create(:user, user_group: ug)
+      login_as(user)
+
+      visit new_service_provider_path
+      expect(page).to have_select('service_provider_user_group_id', selected: ug.name)
```
jessieay: 😄

app/views/service_providers/edit.html.slim
```diff
@@ -1,7 +1,8 @@
 h2 = t('headings.service_providers.edit')
 
 = simple_form_for(service_provider, html: { autocomplete: 'off', role: 'form' }) do |form|
-  == render 'form', form: form
+  - selected_user_group = @service_provider.user_group_id || nil
```
jessieay: you can just write this as `@service_provider.user_group_id` because that will return `nil` if it is `nil` anyway

app/views/service_providers/new.html.slim
```diff
@@ -5,5 +5,6 @@ p = t('links.attr_description_html',
     'https://pages.18f.gov/identity-dev-docs'))
 
 = simple_form_for(@service_provider, html: { autocomplete: 'off', role: 'form' }) do |form|
-  == render 'form', form: form
+  - selected_user_group = current_user.user_group_id || nil
```
jessieay: same comment here...no need to set fallback of `nil`

app/models/user.rb
```diff
@@ -12,6 +12,14 @@ def self.without_other_user_group(user_group_id: nil)
     where(user_group_id: [nil, user_group_id])
   end
 
+  def scoped_user_groups
```
jessieay: this should probably have a unit test....but I'll let it slide for now :)

nickbristow: 😇

app/models/user.rb
```diff
@@ -12,6 +12,14 @@ def self.without_other_user_group(user_group_id: nil)
...
+    if admin?
+      UserGroup.all
+    else
+      UserGroup.where(id: user_group)
```
jessieay: this can be re-written as `user_group`

PR #116, opened 2017-03-29
Hide user group on service provider form if user doesn't have a user group

nickbristow:


app/views/service_providers/_form.html.slim
```diff
@@ -7,7 +7,8 @@
   as: :collection_select,
   collection: current_user.scoped_user_groups,
   include_blank: '',
-  selected: selected_user_group
+  selected: selected_user_group,
+  disabled: !current_user.user_group && !current_user.admin?
```
jessieay: let's throw this logic in a helper method so we can unit test it!

spec/features/service_providers_spec.rb
```diff
@@ -10,6 +10,7 @@
       visit new_service_provider_path
 
       expect(page).to_not have_content('Approved')
+      expect(page).to have_field('User Group', disabled: true)
```
jessieay: if we have a unit test for this behavior, we can remove this expectation :)

spec/models/user_spec.rb
```diff
@@ -65,4 +65,44 @@
...
+    it 'returns collection of users user groups' do
+      group = create(:user_group)
+      user.user_group = group
+      user.save
+
+      expect(user.scoped_user_groups).to eq([group])
+    end
+
+    it 'returns all user groups for admins' do
+      5.times do
```
jessieay: I don't think we need 5 records here - 2 should be enough :)

app/models/user.rb
```diff
@@ -20,6 +20,10 @@ def scoped_user_groups
     end
   end
 
+  def user_group?
```
jessieay: to reveal the intent of this method, I might call it something like `can_select_user_group?` and put it somewhere outside of the user model. Like maybe in a helper? Usually I see view-specific methods in helper modules rather than in ActiveRecord classes.

app/views/service_providers/_form.html.slim
```diff
@@ -7,7 +7,8 @@
   as: :collection_select,
   collection: current_user.scoped_user_groups,
   include_blank: '',
-  selected: selected_user_group
+  selected: selected_user_group,
+  disabled: !current_user.user_group?
```
jessieay: per my naming recommendation above, I also think this would be clearer if it read something like `disabled: !can_edit_user_group?`

spec/models/user_spec.rb
```diff
@@ -65,4 +65,44 @@
       expect(user.scoped_service_providers).to eq([user_sp, group_sp])
     end
   end
+
+  describe '#scoped_user_groups' do
```
jessieay: yay tests!

app/controllers/service_providers_controller.rb
```diff
@@ -1,4 +1,5 @@
 class ServiceProvidersController < AuthenticatedController
+  helper UserGroupHelper
```
jessieay: I don't think you need to include this. think it's available in all views my default....but I could be wrong

PR #117, opened 2017-03-30
Add ability to set logo name when adding SP

jessieay:
**WHY**: Agencies can add logos!

When logo name is invalid:

screen shot 2017-03-30 at 11 56 40 am

When logo name is valid:

screen shot 2017-03-30 at 11 57 40 am

### DIFF
--- a/app/controllers/service_providers_controller.rb
+++ b/app/controllers/service_providers_controller.rb
@@ -96,7 +96,7 @@ class ServiceProvidersController < AuthenticatedController
       :description,
       :friendly_name,
       :issuer,
-      # :logo,
+      :logo,
       :metadata_url,
       :redirect_uri,
       :return_to_sp_url,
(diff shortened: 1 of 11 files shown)


jessieay:
cc @blacktm for input on UX for this

blacktm:
Yea, I agree with you @nickbristow, "Logo" is probably fine. We'll have instructions/context both in our dev docs and in outreach emails to agencies.

blacktm:
Other than that, I think this looks great. I didn't even think to preview the logo in the dashboard (duh) — great idea!

jessieay:
@blacktm @nickbristow good call. Updated at 4f65318

Also updated the preview to show the generic logo svg if no logo is set so they will know what logo will show for them in the SP.

blacktm:
Great, I think good to go then, unless @nickbristow has any last thoughts.

PR #118, opened 2017-04-04
Update dashboard to use new prod name, remove demo

zachmargolis:


app/services/deploy_status_checker.rb
```diff
@@ -5,7 +5,7 @@ class DeployStatusChecker
   Environment = Struct.new(:env, :statuses)
 
   DEPLOYS = [
-    Deploy.new('identity-idp', 'prod', 'https://idp.prod.login.gov'),
+    Deploy.new('identity-idp', 'prod', 'https://secure.login.gov'),
     Deploy.new('identity-sp-rails', 'prod'),
     Deploy.new('identity-sp-sinatra', 'prod'),
     Deploy.new('identity-dashboard', 'prod'),
```
amoose: These can be removed since we don't have any Demo SPs in prod

zachmargolis: Without a URL we just show an empty box, I kept it to keep the layout consistent?

identity_dashboard

PR #120, opened 2017-04-05
Removes demo env from capistrano 

amoose:
### Why
We no longer have a Demo env

### How
Removes Demo from capistrano stages

### DIFF
--- a/config/deploy/demo.rb
+++ b/config/deploy/demo.rb
@@ -1,1 +1,0 @@
-set :bastion_host, 'jumphost.demo.login.gov'


zachmargolis:
I did this as part of https://github.com/18F/identity-dashboard/pull/118 ?

zachmargolis:
Nvm I forgot to remove the deploy file 😬

amoose:
Oh look at that, I didn't even check. I was thinking you were just removing the demo/example SPs. Either way!

PR #123, opened 2017-04-17
Show link when SVG logo is used

nickbristow:
**Why:**   SVG's wont load from an external URL
  Shows link to svg preview when svg is used


jessieay:
I will wait to review until travis is happy :)

PR #124, opened 2017-05-10
Added select for identity protocol to service providers

nickbristow:
Allows users to set SAML or OpenId Connect
  Will change which fields are shown in future PR
identity_dashboard

### DIFF
--- a/app/models/service_provider.rb
+++ b/app/models/service_provider.rb
@@ -4,6 +4,7 @@ class ServiceProvider < ActiveRecord::Base
   belongs_to :user_group
 
   enum block_encryption: { 'aes256-cbc' => 1 }
+  enum identity_protocol: { openid_connect: 0, saml: 1 }
 
   validates :issuer, presence: true, uniqueness: true
   validates :agency, presence: true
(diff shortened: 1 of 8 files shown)


app/models/service_provider.rb
```diff
@@ -4,6 +4,7 @@ class ServiceProvider < ActiveRecord::Base
   belongs_to :user_group
 
   enum block_encryption: { 'aes256-cbc' => 1 }
+  enum identity_protocol: { 'OpenID Connect' => 0, 'SAML' => 1 }
```
zachmargolis: Should we make these more slug-like and then add translations for them to get the pretty-printed names? ex `openid_connect` and `saml` as symbols?

monfresh: Yes, please, to Zach's comment.

nickbristow: @monfresh @zachmargolis like so ?
```
enum identity_protocol: { openid_connect: 0, saml: 1 }
```

monfresh: Yes sir!

zachmargolis: 💯 yay thanks!

PR #126, opened 2017-05-15
Show appropriate fields based on id protocol

nickbristow:
**Why**
Users were filling out unneccessary fields when creating sp's
This eliminates confusion by hiding those fields


blacktm:
This works great.

PR #127, opened 2017-05-18
Automatically publish service providers upon save or update

nickbristow:
**Why**
Streamlines the process and eliminates an uneccesary extra step


app/controllers/service_providers_controller.rb
```diff
@@ -59,9 +59,18 @@ def save_service_provider(initial_action)
     service_provider.save!
     flash[:success] = I18n.t('notices.service_provider_saved', issuer: service_provider.issuer)
     notify_users(service_provider, initial_action)
+    publish_service_providers
     redirect_to service_provider_path(service_provider)
   end
 
+  def publish_service_providers
+    if ServiceProviderUpdater.new.ping
```
blacktm: I wonder if `ServiceProviderUpdater` should just expose class methods. Seems wasteful to instantiate.

PR #128, opened 2017-05-19
Changed documentation link on SP form to open in a new tab

nickbristow:
**Why**
Users who clicked the documentation link
would be directed away from the form

Also added documentation link to the form partial.
Link shows on edit and new forms now.


app/views/service_providers/_form.html.slim
```diff
@@ -1,3 +1,7 @@
+p = t('links.attr_description_html',
+    href: link_to(t('links.attr_description_href'),
+    'https://pages.18f.gov/identity-dev-docs', target: :_blank))
```
zachmargolis: this is out of date now! it's https://developers.login.gov

nickbristow: good catch

PR #129, opened 2017-05-31
Support multiple redirect_uri values

zachmargolis:
**Why**: Better OIDC support

First step of multiple, eventually will follow up with a PR to drop the old column once we have everything running off the new column

### DIFF
--- a/db/migrate/20170531194154_add_multiple_redirect_uris.rb
+++ b/db/migrate/20170531194154_add_multiple_redirect_uris.rb
@@ -1,0 +1,15 @@
+class AddMultipleRedirectUris < ActiveRecord::Migration
+  def up
+    add_column :service_providers, :redirect_uris, :json
+
+    execute <<-SQL
+      UPDATE service_providers
+      SET redirect_uris=array_to_json(ARRAY[redirect_uri])
+      WHERE length(redirect_uri) > 0
+    SQL
+  end
+
+  def down
+    remove_column :service_providers, :redirect_uris, :json
+  end
+end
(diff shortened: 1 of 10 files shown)


db/migrate/20170531194154_add_multiple_redirect_uris.rb
```diff
@@ -0,0 +1,15 @@
+class AddMultipleRedirectUris < ActiveRecord::Migration
+  def up
+    add_column :service_providers, :redirect_uris, :json
```
monfresh: Why JSON and not Array?

zachmargolis: I was matching the style of the codebase already, there's a pre-existing JSON column but no other array columns:

```
identity-dashboard:master> git grep array -- db/migrate/
identity-dashboard:master> git grep json -- db/migrate/
db/migrate/20170117222117_add_attribute_bundle.rb:    add_column :service_providers, :attribute_bundle, :json
```

monfresh: I wouldn't consider choice of column types a style decision. Wouldn't it make sense to choose the right column type for the job?

zachmargolis: I think it's acceptable as is, how strongly do you feel that it should be changed to array?

monfresh: Not strongly.

PR #130, opened 2017-06-01
Remove foreman from Gemfile

jgrevich:
__Why__

* The older version is causing builds to fail.

```
Downloading foreman-0.83.0 revealed dependencies not in the API or the lockfile (thor (~> 0.19.1)).
Either installing with `--full-index` or running `bundle update foreman` should fix the problem.
```

__How__

* Update foreman as the error message suggests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,6 @@ gem 'devise'
 gem 'dotenv-rails'
 gem 'enum_help'
 gem 'figaro'
-gem 'foreman', require: false
 gem 'httparty'
 gem 'jquery-rails'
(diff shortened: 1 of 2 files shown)


monfresh:
I just ran into this issue as well and just noticed this PR. We actually don't want foreman in the Gemfile at all. The README explicitly says to make sure you don't install it via the Gemfile. @jgrevich can you please update this PR to remove foreman from the Gemfile? Thanks!

monfresh:
Foreman is installed and updated via our `bin/setup` script.

jgrevich:
All set. Thanks for pointing that out @monfresh, I forgot about that.

PR #134, opened 2017-06-05
Allow users to be in multiple groups

nickbristow:
**Why**
Users should be able to join multple groups to
get access to multiple sets of service providers
as needed.

Renamed user_groups to groups and now use user_groups
as the join table between user and groups.


nickbristow:
So this PR got rather large as I renamed user_groups to just groups, and used user_groups as the join table between...users and groups. This resulted in changing every file that referenced user_group to just group. 

@blacktm I haven't written a rake task to put users back in their groups yet, let me know if you think we need one, or want to do it by hand on prduction.

zachmargolis:
@nickbristow shouldn't we be able to migrate user groups as we go without having to do anything by hand?

app/views/groups/index.html.slim
```diff
@@ -0,0 +1,44 @@
...
+          - group.users.each do |user|
+            li
+              = user.email
+      td
+        = link_to 'Edit',
+          edit_group_path(group),
+          'aria-label' => t('links.aria.edit', name: group.name)
+      td
+        = link_to 'Delete',
+        group_path(group),
```
zachmargolis: nit: indent this to match `link_to`?

nickbristow:
@zachmargolis yeah, i've left the direct connection intact for now. Would a temporary rake task be best for that?

zachmargolis:
@nickbristow I've seen rails migrations that do this? Easier to remember to just run migrations rather than try to remember to run a rake task in all 5-6 of our environments?

I'll have to look at the table structures more closely, but I think this should be fairly simple to automate since we're turning a 1:1 relation into a 1:many?

nickbristow:
@zachmargolis easy enough to do in a migration, I've just read a bit about referencing models and manipulating data in migrations is frowned upon. 

I can do it if you ok that.

zachmargolis:
I'm ok with that! I would also recommend writing the data migrations in raw SQL (because the various activerecord classes that exist to back the tables may not exist at each point in time, this bit us pretty bad in IDP)

PR #135, opened 2017-06-07
Add OIDC sinatra example app to dashboard

zachmargolis:
identity_dashboard


app/services/deploy_status_checker.rb
```diff
@@ -47,6 +18,10 @@ def status_class
       end
     end
 
+    def short_name
+      app.gsub('identity-', '')
+    end
```
monfresh: Add a test for this?

zachmargolis: Done! eff7c088ae1259f299551b0af844a70ed805d017

config/status_checks.yml
```diff
@@ -0,0 +1,88 @@
...
+    env: 'prod'
+    host: 'https://secure.login.gov'
+  - app: 'identity-sp-rails'
+    env: 'prod'
+  - app: 'identity-sp-sinatra'
+    env: 'prod'
+  - app: 'identity-dashboard'
+    env: 'prod'
+  - app: 'identity-openidconnect-sinatra'
+    env: 'prod'
```
zachmargolis: Configs without a host are not checked and shown as unavailable, so these are being "excluded" but we keep them in the UI for consistency (that's pre-existing behavior)

monfresh: Gotcha. Thanks for the explanation.

PR #136, opened 2017-06-20
Validate format of issuer in ServiceProvider model

jmhooper:
**Why**: This verifies that service providers use a clear and consistent
schema for names. Since this is a validation and not a database
constraint,

Add a hint about issuer format to the service provider form

**Why**: Users creating or updating service providers need to know what
the valid format is. This commit adds a hint to the form that describes
the proper format

Update the issuer hint and value when a user switches between SAML and
OpenID connect on the service provider form

**Why**: When the user updates between SAML and OpenID connect, they
will want to modify the format of their issuer. This commit modifies
the existing Javascript event listener to handler that step for them.


app/assets/javascripts/app/service-provider-form.js
```diff
@@ -19,13 +19,36 @@ $(function(){
...
+      case 'openid_connect':
+        hint_text = hint.html().replace('SAML:2.0', 'openidconnect');
+        issuer_input_value = issuer_input.val().replace('SAML:2.0', 'openidconnect');
+        break;
+      case 'saml':
+        hint_text = hint.html().replace('openidconnect', 'SAML:2.0');
+        issuer_input_value = issuer_input.val().replace('openidconnect', 'SAML:2.0');
+        break;
+    }
+    hint.html(hint_text);
```
zachmargolis: Do we need `.html()` here or can we use plain `.text()`?

spec/factories/service_providers.rb
```diff
@@ -1,7 +1,7 @@
 FactoryGirl.define do
   factory :service_provider do
     sequence(:friendly_name) { |n| "test-service_provider-#{n}" }
-    sequence(:issuer) { |n| "test-service_provider-#{n}" }
+    sequence(:issuer) { |n| "urn:gov:gsa:SAML:2.0.profiles:sp:sso:DEPT:APP-#{n}" }
```
zachmargolis: if the only bits that change are `DEPT` and `APP`, what if we just make those the user-editable parts in the form and generate the issuer from that?

jmhooper: Are you thinking about changing the `issuer` to `issuer_department` and `issuer_agency` in the model? I'm not sure that will work with the ServiceProviders that have been created with an issuer that does not match the format.

If we want to keep the schema that way it is and only change it in the UI and build the issuer in the controller, then the challenge is:

1. How to actually lay this out in the UI. We'd want a control that communicates that the `DEPT` and `APP` strings are components of a uid.
2. Parsing the `DEPT` and `APP` strings out of the `issuer`, especially for existing service providers who have what we'd now describe as a malformed issuer.

zachmargolis: I would aim to only change in the UI and leave the data model the same.

- For new issuers and ones that parse correctly, we could break out two fields.
- For old issuers that do not parse correctly we could just fall back and render a single field?

Just an idea? I don't feel that strongly about it either way

jmhooper: Hmm, new issuers will need to change anyway. Perhaps a message saying something to the end of `Issuer is invalid and needs to be updated` in the UI if the department and agency strings can't be parsed out of the issuer string?

zachmargolis: the issuer is our primary key in the IDP, I really don't think we should force SPs to change since there would be no continuity? That would cause use to generate new UUIDs for existing users for the SPs?

This is the dashboard, so it only affects lower envs, maybe that's OK? Happy to chat more about this in the orignal issue if needed

amoose: > I really don't think we should force SPs to change since there would be no continuity? 

I don't see a way around this; if the current format does not conform to what we have prescribed, they should update it, regardless if we enforce it via app validations or face to face communication.

app/models/service_provider.rb
```diff
@@ -1,17 +1,33 @@
...
+  validates :issuer_department, presence: true
+  validates :issuer_app, presence: true
   validates :agency, presence: true
 
   before_validation(on: %i(create update)) do
     self.attribute_bundle = attribute_bundle.reject(&:blank?) if attribute_bundle.present?
   end
+  before_validation :build_issuer
+
+  def issuer_department
```
jmhooper: I'm not the biggest fan of bolting this onto the model, but adding a Form object to handle this started to feel heavy handed. I did the best I could to keep most of the logic in service objects.

app/services/service_provider_issuer_builder.rb
```diff
@@ -0,0 +1,25 @@
...
+  ISSUER_TEMPLATE = 'urn:gov:gsa:{protocol}.profiles:sp:sso:{department}:{app}'.freeze
+
+  attr_reader :service_provider
+
+  def initialize(service_provider)
+    @service_provider = service_provider
+  end
+
+  def build_issuer
+    ISSUER_TEMPLATE.gsub('{protocol}', protocol_substring).
```
zachmargolis: If we use the `%{}` syntax we could just use the built-in % interpolation!

```ruby
ISSUER_TEMPLATE = 'urn:gov:gsa:%{protocol}.profiles:sp:sso:%{department}:%{app}'.freeze

def build_issuer
  ISSUER_TEMPLATE % {
    protocol: protocol_string,
    department: service_provider.issuer_department,
    app: service_provider.issuer_app
  }
end
```

jmhooper:
I changed this up so that instead of entering the `issuer` string, the user enters the department and app substrings. From there, the protocol substring is inferred from the protocol the service provider is using.

There's some Javascript to make sure the actual value of the issuer is displayed in the fronted as the user is updating the substrings.

This should still not force existing SPs to change their issuer until something on them is updated.

app/services/service_provider_issuer_parser.rb
```diff
@@ -0,0 +1,15 @@
...
+  attr_reader :issuer
+
+  def initialize(issuer)
+    @issuer = issuer
+  end
+
+  def parse
+    match_data = ServiceProvider::ISSUER_FORMAT_REGEXP.match(issuer) || []
+    {
+      department: match_data[2],
```
zachmargolis: Named captures might make this easier:

```ruby
ISSUER_FORMAT_REGEXP = /\Aurn:gov:gsa:(?<protocol>SAML:2\.0|openidconnect)\.profiles:sp:sso:(?<department>.+):(?<app>.+)\z/

match_data = ISSUER_FORMAT_REGEXP.match("urn:gov:gsa:openidconnect.profiles:sp:sso:TEST:BAR")

match_data[:department]
# "TEST"
```

app/views/service_providers/_form.html.slim
```diff
@@ -16,8 +16,12 @@ p = t('links.attr_description_html',
 .common-fields
   = form.input :friendly_name
   = form.input :description
-  = form.input :issuer,
-    input_html: { value: form.object.issuer || 'urn:gov:gsa:SAML:2.0.profiles:sp:sso:' }
+  .mb2.service_provider_issuer
+    label Issuer
+    p#service_provider_issuer
+      = form.object.issuer || "Enter department and app to generate uid"
```
zachmargolis: We've moved to i18n strings in this app, let's move this to `en.yml` ?

app/services/service_provider_issuer_builder.rb
jmhooper: The linter got mad about the `a % b` syntax, but I was able to use `Kernel#format` here.

jmhooper:
@zachmargolis: All good feedback. I've pushed some commits to address those points.

jmhooper:
Had a chat with @zachmargolis. We discussed making the issuer readonly and then only running this validation on create. That allows old SPs to keep their current issuers, but forces new issuers to use the new format.

SPs would no longer be able to update their issuers. His consideration there was that, since the issuer is a PK in several of our systems, it may be more appropriate to create a new issuer instead.

jmhooper:
I've changed this PR up so that now the issuer is a readonly attribute. In addition, the format validation for the issuer is only run on create, which will allow service provider with malformed issuers to exist in peace, but require new service providers to use the correct format.

The issuer agency and issuer department substring inputs are only displayed in the form for new records, and are not available when updating a record.

Also, I fixed the setup script, which ran into a problem in the `dev:prime` rake task where the initial SP it was creating used a UUID for an issuer instead of our new fancy pants string.

app/models/service_provider.rb
```diff
@@ -1,4 +1,9 @@
 class ServiceProvider < ActiveRecord::Base
+  ISSUER_FORMAT_REGEXP = /\Aurn:gov:gsa:(?<protocol>SAML:2\.0|openidconnect)\.profiles:sp:sso:(?<department>.+):(?<app>.+)\z/
```
zachmargolis: looks like we're getting a rubocop error here still

some options:
1. rubocop ignore comment
2. `/x` regex mode to make it a multi-line regex (https://ruby-doc.org/core-2.1.1/Regexp.html#class-Regexp-label-Free-Spacing+Mode+and+Comments)

app/assets/javascripts/app/service-provider-form.js
```diff
@@ -19,13 +21,44 @@ $(function(){
     }
   }
 
+  var issuer_protocol_name = function(id_protocol){
+    switch(id_protocol){
+      case 'openid_connect':
+        return 'openidconnect';
+        break;
```
zachmargolis: Looks like code climate is mad at the `break` after a `return` because it's unreachable (since the `return` leaves the function before the `break`, can we remove these?

lib/tasks/dev.rake
```diff
@@ -9,10 +9,9 @@ if Rails.env.development? || Rails.env.test?
       user = create(:user, email: 'user@example.com')
 
       dashboard_saml_config = Saml::Config::SETTINGS
-      uuid = SecureRandom.uuid
```
zachmargolis: I just ran this locally and got an error because `uuid` (used in the `metadata_url`) is now undefined:

```
NameError: undefined local variable or method `uuid' for main:Object
```

PR #137, opened 2017-07-18
Downgrades bummr gem to 0.1.8

amoose:
#### Why
Version 0.2.0 has been **yanked** 🔪 
https://rubygems.org/gems/bummr/versions/0.2.0

#### How
Pins bummr gem version to 0.1.8

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,7 +33,7 @@ group :deploy do
 end
 
 group :development do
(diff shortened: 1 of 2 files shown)


monfresh:
version 0.2.1 is now available. I will close this and submit a new PR that updates gems with bummr.

PR #146, opened 2017-11-15
Add ACS url to SAML config

jmhooper:
**Why**: So that we can be certain we're comparing against the correct
ACS url when validating the SAML subject confirmation.


blacktm:
We actually just switched the dashboard over to use OIDC instead of SAML (#152).

PR #147, opened 2017-12-26
Update gems with bummr

monfresh:


.rubocop.yml
```diff
@@ -154,7 +154,6 @@ Style/IfUnlessModifier:
   Description: Favor modifier if/unless usage when you have a single-line body.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#if-as-a-modifier
   Enabled: true
-  MaxLineLength: 100
```
tbaxter-18f: Why was this removed?

monfresh: It has been deprecated by Rubocop. This cop now automatically inherits the max line length from the `Metrics/LineLength` cop.

brodygov:
Dashboard deploys are failing on servers with these changes with a message about uglifier.

PR #151, opened 2018-03-01
Use the U.S. Web Design System

blacktm:
**Why**: We currently have a mashup of older USWDS assets, Basscss, and custom styles. This replaces all of it with the latest USWDS release and adds only minor style adjustments needed by the dashboard. Also, we opted for using compiled assets available in the `public/` dir, instead of the Sass source assets.

We're doing a more thorough update of the view in a later PR.


s2t2:
@blacktm since there are so many file changes I'm focusing my review on user experience.

I'm running the app locally and wondering if these screenshots match your expectations:

screen shot 2018-03-06 at 5 44 06 pm


screen shot 2018-03-06 at 5 49 44 pm


screen shot 2018-03-06 at 5 50 04 pm

screen shot 2018-03-06 at 5 50 14 pm

Also I noticed I was unable to edit a service provider's "User Group" using the form (maybe a separate issue from this PR).

blacktm:
Yep, that's looks right, largely should match the visual experience of [developers.login.gov](https://developers.login.gov). In this PR, it's only the old assets being removed, and the new static assets added to the `public/` dir, a [direct download from USWDS](https://designsystem.digital.gov/getting-started/download/) making it easy to update with future releases. This is the same thing we do [for the dashboard](https://github.com/18F/identity-dev-docs/tree/master/assets). Ideally, we would link to a gov-hosted CDN, but in the absence of that, this is probably the next best route. In another PR, I'm going to work on the views specifically, clean those up.

PR #155, opened 2018-03-09
Make creating and managing test apps easier on agencies

blacktm:
- Disables validation of issuer at creation, since some agencies are
unable to easily change it
- Makes the friendly name required
- Removes the `agency` field from the form and allows it to be null,
since it's basically useless in our integration env
- Make new entries active by default
- Make the user group description field optional


blacktm:
Added comments to [service-provider-form.js](https://github.com/18F/identity-dashboard/pull/155/files#diff-481744ba8584819ae617aa3b185d81f2) and [models/service_provider.rb](https://github.com/18F/identity-dashboard/pull/155/files#diff-772780d5c66d5ab6911a0494e8a0f774) regarding the temporary disabling of the issuer field.

PR #156, opened 2018-03-09
Provide more redirect URI fields in SP form

blacktm:
**Why**: The current UI only shows one redirect URI field, so this adds a button to append more.


blacktm:
We actually already have a `app/assets/javascripts/app/service-provider-form.js` file (duh), so I moved it there. Also, forgot to wrap both the `new` and `edit` forms with the `service-provider-form` class, so that's now fixed also.

PR #159, opened 2018-03-14
Update IdP and dashboard URLs

blacktm:


mark-from-usds:
I don't think this will work without https://github.com/18F/identity-devops/pull/791 from another repo. Otherwise "idp_url" won't be defined.

blacktm:
@mark-from-usds Ah yes, you're right. There's currently not a place where the raw IdP URL is exposed.

mark-from-usds:
OK I've merged that PR into devops master.

blacktm:
@mark-from-usds Awesome, ready to merge this then? I've tested locally, btw.

PR #167, opened 2018-04-28
Remove delayed_job

monfresh:
**Why**: It's not used by the app


monfresh:
Note that there is a delayed_job table in the DB that will be removed in a separate PR just in case there are migration issues.

PR #168, opened 2018-04-28
Update to version 2 of Code Climate

monfresh:


monfresh:
This PR is to make code quality checks consistent with our other repos. Most of the changes are from Rubocop's autocorrect feature. There are still other issues to fix, but those will be addressed in a seaparate

PR #175, opened 2018-07-12
Specify Rails version in migrations

### LINKED ISSUE #174: Migrations are broken in Rails 5
```
StandardError: Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

class AddUsersAndApplications < ActiveRecord::Migration[4.2]
```

monfresh:
**Why**: Rails 5 requires this when running migrations.
Fixes #174


brodygov:
Thanks

PR #178, opened 2018-08-12
Fix grammar typo on app update form

konklone:
This pluralizes a word and adds a `that` to make it more clear.


konklone:
@monfresh Do you know why the required statuses aren't completing here? Are some out of date or something?

monfresh:
Sometimes, Code Climate gets stuck. I can merge this. Also, we should look into ways to skip CI for certain PRs. Text changes like this one don't require CI. I've seen other GitHub projects add something like `[skip CI] ` to their PR title. I'm not sure if that's a magic string that works for all GitHub PRs and CI services.

konklone:
> Also, we should look into ways to skip CI for certain PRs. Text changes like this one don't require CI.

I'm not sure that's the right takeaway here -- some of the tests for this repository specifically look at copy, and at least one of the tests verifies that some copy is in place because to change it would trigger New Relic to think the site is down.

PR #182, opened 2018-09-21
Allow disabling SAML assertion encryption.

brodygov:
We support this in the IDP, but there's no way to configure it from the
dashboard. Allow block_encryption to be "none", since encryption of
SAML assertions has little to no security value for PII that isn't
sensitive.

This requires adding a suffix to the block_encryption model enum,
because in rails you can't create an enum with value "none" because it
clashes with the existing class method `.none`.


brodygov:
This is deployed and appears to be working at https://dashboard.brody.identitysandbox.gov

PR #186, opened 2018-11-29
LG-822 Order applications and display a creation timestamp

stevegsa:
We need an easy way to see when a new application is added to the sandbox.


stevegsa:
Screenshots:

screen shot 2018-11-28 at 9 20 25 pm

PR #191, opened 2018-12-15
LG-838 Admins can indicate which apps are live

stevegsa:
Admins can indicate which apps are live by editing the production_issuer for any app.


stevegsa:
Screenshots:
screen shot 2018-12-15 at 12 32 53 am

screen shot 2018-12-15 at 12 33 36 am

screen shot 2018-12-15 at 12 34 33 am

app/controllers/service_providers_controller.rb
```diff
@@ -114,6 +114,7 @@ def service_provider_params
       :sp_initiated_login_url,
       :group_id,
       :identity_protocol,
+      :production_issuer,
```
jmhooper: Looks like there's a mass assign issue here. A non-admin user can add a value for `production_issuer` to the request and set the production issuer.

jmhooper: If we are going to allow them to modify it, then let's not hide the field. If we aren't going to allow them to modify if, then let's follow through in the controller and build some tests around it.

stevegsa: I opted to not let them modify it.  It would be too confusing for them and they might get the information wrong.

PR #192, opened 2019-01-04
LG-838 Admins can indicate which apps are live

stevegsa:
Admins can indicate which apps are live by editing the production_issuer for any app. There is no standardized naming for issuer so there is no way to automate the identification of which apps are in production. Rather than simply providing a boolean for the app, a production issuer will allow us to provide a reference to the prod issuers.


stevegsa:
This was an attempt to unstick the other LG-838.  Closing...

PR #198, opened 2019-03-20
Add IAL attribute to service providers.

brodygov:
This is a dependency for https://github.com/18F/identity-idp/pull/2672

The IDP will soon start enforcing validation such that IAL1 service
providers can only request IAL1 attributes, not sensitive attributes.

By default the IDP will consider service providers to be at IAL1. So in
order to avoid any service providers that are relying upon IAL2/LOA3
data, we have to add a corresponding attribute to service providers in
the dashboard.

### DIFF
--- a/app/controllers/service_providers_controller.rb
+++ b/app/controllers/service_providers_controller.rb
@@ -104,6 +104,7 @@ class ServiceProvidersController < AuthenticatedController
       :description,
       :friendly_name,
       :group_id,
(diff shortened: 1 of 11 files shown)


brodygov:
Tests are still broken, but I'm tagging people to solicit feedback.

brodygov:
Screen Shot 2019-03-20 at 6 38 53 PM

PR #201, opened 2019-03-27
Minor tweak to form to use cert instead of key

amoose:
Using "public key" is a little misleading since we're actually asking for the public certificate. Added more clear language to state this.


amoose:
I agree with you, Andy. Next phase of updates!

PR #202, opened 2019-03-29
LG-1081 - Updates validation for Issuer

amoose:
The flood gates were opened when we commented out the validations on Issuer

Completes: https://cm-jira.usa.gov/browse/LG-1081

This PR re-enables basic validation on Issuer:

Re-enables format validation on Issuer to block whitespace
Removes commented out and unused code around Issuer builder and parser
Removes unused specs and code


amoose:
Still cleaning out some cruft. 🔪

amoose:
I will follow up with more code coverage separately.

app/models/service_provider.rb
```diff
@@ -72,7 +61,7 @@ def self.possible_attributes
     ]
     Hash[*possible.collect { |v| [v, v] }.flatten]
   end
-  # rubocop:ensable MethodLength
```
jmhooper: nice

amoose: 💃

stevegsa:
the @amoose is back!

PR #203, opened 2019-05-30
LG-1035 - Updates formatting for redirect URIs

amoose:
Forces URIs to be lowercase. That's it. 

This solves issues with CSP errors where the redirect is to a lowercase URI despite the whitelist containing CamelCase. The root cause is the http-level redirect uses lowercase (or is interpreted as lowercase per the spec) but the CSP contains the case-sensitive URI.


amoose:
I'll follow up with a PR for the IdP

stevegsa:
Is it possible there are clients that are case sensitive?

PR #209, opened 2019-07-03
Test improvements

jennywagg:
While familiarizing with the identity-dashboard project, I made some minor improvements and added some tests to improve coverage.


spec/policies/user_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  let(:app)        { build(:service_provider, user: owner) }
+
+  permissions :create? do
+    it 'allows any user to create' do
+      expect(ServiceProviderPolicy).to permit(user, app)
+    end
+  end
+
+  permissions :edit? do
+    it 'allows owner or admin to edit' do
```
solipet: I realize the other policy spec is structured this way, but let's keep one `expect` per `it`, e.g.,
```
  it 'allows owner to edit' do...
  it 'allows admin to edit' do...
  it 'does not allow other users to edit' do...
```

PR #210, opened 2019-07-05
Adds Group 'show' view

amoose:
## Why:

We don't have a nice way to view details about a group.

## How:

Adds a group 'show' view. 

### Also:

Re-enables the group feature spec. 
Enables truncation for the database cleaner method. 

#### Screenshot

Screen Shot 2019-07-05 at 4 19 24 PM


amoose:
I guess I should just fix the sub-nav. Might as well break it out in to a partial for when the design system lands. 🛬

app/views/groups/show.html.erb
```diff
@@ -0,0 +1,24 @@
+<h1>Group details</h1>
+
+<!-- TODO: FIX NAV! eek -->
```
amoose: 🐽

amoose:
I'm going to hold off on improving the subnav until the design system lands. We'll break it out properly when we've adopted the new UI and began utilizing it.

app/views/shared/_subnav.html.erb
```diff
@@ -0,0 +1,12 @@
+<!-- Reminder: never use user-input for current_item -->
+<% item = current_item.class.to_s.downcase %>
+
+<% index_path = controller.send(item.pluralize + '_path') %>
+<% edit_path = controller.send('edit_' + item + '_path') %>
+<% new_path = controller.send('new_' + item + '_path') %>
```
amoose: This is... not pretty. We'll address it after we implement the design systemm.

amoose: I might as well remove it. :thinkin'

PR #211, opened 2019-07-05
Adds access request form to unauthorized page

amoose:
Why:

The UX is not great and we want to better leverage the CRM to handle inbound leads. 

How:

Embeds our CRM's request access form to the unauthorized page.

Also:

Makes the app name more consistent (so it's easier to change in the future!).

<hr>
Before:
Screen Shot 2019-07-05 at 5 07 56 PM

After:
Screen Shot 2019-07-05 at 4 43 10 PM

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -4,7 +4,7 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>login.gov Dashboard</title>
+    <title><%= Rails.application.config.app_name %></title>
     <link rel="stylesheet" href="/assets/uswds/css/uswds.min.css">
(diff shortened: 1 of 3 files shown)


mguzman-gov:
@amoose please see my suggested edits below

_Change to the intro:_
"We were unable to find an account matching your email address in the Partner Dashboard."

_Edits for capitalization and one missing comma shown in bold_
"Federal **a**gency **n**ame
**l**ogin.gov is currently limited to servicing **f**ederal **a**gencies. We love working with private companies**,** but you must be under contract with a federal agency in order to use **l**ogin.gov. 

_The following text feels out of place there. Suggesting removing it and leaving the rest of the line_
"We have integrated 15+ agencies,"

amoose:
Thanks for the feedback, @mguzman-gov! Updated the copy to include your edits. 

Screen Shot 2019-07-08 at 2 58 19 PM

PR #212, opened 2019-07-11
LG-1457: saml redirect uris

solipet:
Allow Service Providers using the SAML protocol to specify multiple redirect uris to be included in the SAML response CSP headers.

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -1,6 +1,6 @@
 class GroupsController < ApplicationController
   before_action -> { authorize Group }
-  before_action :find_group, only: %i(show edit update destroy)
+  before_action :find_group, only: %i[show edit update destroy]
 
   def new
     @group = Group.new
--- a/app/views/service_providers/_form.html.erb
+++ b/app/views/service_providers/_form.html.erb
@@ -32,10 +32,8 @@
 
 <%= form.input :return_to_sp_url, input_html: { class: 'h4p4' }, label: "<b>Return to App URL</b><br>The applications URL which login.gov provides to users when they wish to go directly to the app's site or cancel out of authentication. For example: <code>https://app.agency.gov</code>".html_safe %>
 
-<div class="oidc-fields">
-  <%= form.input :redirect_uris, label: "<b>Redirect URIs</b> — <i>OpenID Connect only</i><br>One or more URIs that login.gov will redirect to after authentication. These can be web URLs (public, internal, or localhost) or a custom scheme to support native applications, for example: <code>gov.agency.app://result</code>".html_safe %>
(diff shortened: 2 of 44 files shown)


solipet:
That's what I get for fixing all rubocop complaints... 3 code climate errors. The nil checks are part of a Policy that check if a user is an admin, so I think those are fair.  How to mark errors as `ignore` or `won't fix`?

app/controllers/groups_controller.rb
```diff
@@ -1,6 +1,6 @@
 class GroupsController < ApplicationController
   before_action -> { authorize Group }
-  before_action :find_group, only: %i(show edit update destroy)
+  before_action :find_group, only: %i[show edit update destroy]
```
amoose: group view action pending ;)

app/views/service_providers/_form.html.erb
```diff
@@ -33,10 +33,16 @@
...
 
 <div class="oidc-fields">
-  <%= form.input :redirect_uris, label: "<b>Redirect URIs</b> — <i>OpenID Connect only</i><br>One or more URIs that login.gov will redirect to after authentication. These can be web URLs (public, internal, or localhost) or a custom scheme to support native applications, for example: <code>gov.agency.app://result</code>".html_safe %>
-  <button id="add-redirect-uri-field" class="usa-button" type="button">Add another URI</button>
+  <%= form.label :redirect_uris, "<b>Redirect URIs</b> — <i>OpenID Connect only</i><br>One or more URIs that login.gov will redirect to after authentication. These can be web URLs (public, internal, or localhost) or a custom scheme to support native applications, for example: <code>x-example-app:/result</code>".html_safe %>
 </div>
 
+<div class="saml-fields">
+  <%= form.label :redirect_uris, "<b>Redirect URIs</b> — <i>SAML only</i><br>One or more URIs that login.gov will include in the SAML response CSP headers to allow multiple targets for redirection. These can be web URLs (public, internal, or localhost) or a custom scheme to support native applications, for example: <code>x-example-app:/result</code>".html_safe %>
```
amoose: I'd like the label to be "_Additional_ Redirect URIs" 

Also, these hints `x-example-app:/result` should be `gov.agency.app://result`.

amoose: Remove "OpenID Connect only" and "SAML only" from the label, too.

app/views/service_providers/show.html.erb
```diff
@@ -69,12 +69,10 @@
       <td><%= service_provider.return_to_sp_url %></td>
     </tr>
 
-    <% if service_provider.identity_protocol == 'openid_connect' %>
     <tr>
       <th scope="row">Redirect URIs</th>
```
amoose: label should match the form: "_Additional_ Redirect URIs"

PR #217, opened 2019-07-22
LG-1410: removes all service_provider validation...

solipet:
and defers to the identity_validations gem.

Why:

It was found that, over time, the validation on ServiceProvider varied between identity-idp and identity-dashboard. 

How:

To ensure that the validations are consistent between repos, both will start using the new  `ServiceProviderValidation` found in the new `identity_validations` gem.


solipet:
This PR is dependent on the new [identity-validations](https://github.com/18F/identity-validations/pull/1) gem and may be reviewed/merged once that gem is reviewed/merged.

PR #218, opened 2019-07-29
LG-1296 service providers can specify agency id through group

jennywagg:


config/application.rb
```diff
@@ -35,5 +35,7 @@ class Application < Rails::Application
     # Make sure the sum of this value and session_timeout_warning_seconds
     # is a multiple of 60 seconds.
     config.session_check_frequency         = 60
+
+    config.agencies = YAML.load_file 'config/agencies.yml'
   end
 end
```
amoose: This is fine for now, but we should think about how to pull this from the IdP in the future.

PR #219, opened 2019-07-30
LG-1594 replace links with buttons

jennywagg:


karlaturcios:
@solipet  if it passes your review, please merge into dashboard1335

PR #228, opened 2019-08-19
Adds NewRelic error handling to SP updater

amoose:
**Why**

Errors with the SP updater do not bubble up to NewRelic

**How**

Adds NewRelic error handling for failed updates so we can properly monitor


app/services/service_provider_updater.rb
```diff
@@ -1,7 +1,13 @@
...
+  # :reek:TooManyStatements
   def self.ping
     resp = HTTParty.post(idp_url, headers: token_header)
-    resp.code == 200
+    status_code = resp.code
+    return status_code if status_code == 200
+
+    handle_error(status_code)
+  rescue StandardError => error
+    handle_error(error.msg)
```
amoose: there's probably a better way to write this? 😴

app/services/service_provider_updater.rb
```diff
@@ -12,5 +18,11 @@ def idp_url
     def token_header
       { 'X-LOGIN-DASHBOARD-TOKEN' => Figaro.env.dashboard_api_token }
     end
+
+    def handle_error(status)
+      ::NewRelic::Agent.notice_error("ServiceProviderUpdater failed with "\
```
solipet: My understanding is that `notice_error` takes an error, not just a string.  I can run with this since I just did similar on the push notification failures.

app/services/service_provider_updater.rb
solipet: @amoose check out my updates...

PR #230, opened 2019-08-20
LG-1641: example push_notification endpoint to test against IDP

solipet:
This allows the dashboard to receive push notifications from the IDP and print the deleted user's details to STDOUT. It's only active in `development` so I thought it would be useful to keep around.


solipet:
Bah, nm, I don't think this is worth supporting with tests...

PR #236, opened 2019-08-29
LG-1273 Add help text fields to sp form, add help text model and associate with service provider

jennywagg:
See: https://github.com/18F/identity-idp/pull/3237


spec/models/help_text_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
```
solipet: Need to write the spec or delete this. Could probably test that sanitization is actually happening...

PR #238, opened 2019-09-04
LG-1815 Generate yaml blob for sp

jennywagg:
**Why:** So that anyone who wants to update a Service Provider in the IdP can view a formatted YAML text blob in the dashboard on that Service Provider's details page and be able to paste it directly into the IdP service providers configuration.

### DIFF
--- a/app/helpers/service_provider_helper.rb
+++ b/app/helpers/service_provider_helper.rb
@@ -1,4 +1,18 @@
+#:reek:UtilityFunction
 module ServiceProviderHelper
+  SP_PROTECTED_ATTRIBUTES = %w[
+    id
+    created_at
+    updated_at
+    user_id
+    description
+    approved
+    active
+    group_id
+    identity_protocol
+    production_issuer
+  ].freeze
+
   def sp_logo(file_name)
     file = file_name || 'generic.svg'
     if file.end_with?('.svg')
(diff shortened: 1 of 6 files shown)


app/views/service_providers/show.html.erb
```diff
@@ -100,6 +100,12 @@
         <%= image_tag service_provider.active? ? 'img/alerts/success.svg' : 'img/alerts/error.svg', height: '27', width: '27', :class =>'margin-bottom-neg-105' %>
       </td>
     </tr>
+    <tr>
+      <th scope="row">Service Provider config as YAML</th>
+      <td>
+        <%= render 'service_provider_yaml' %>
+      </td>
+    </tr>
```
amoose: Let's restrict this to only admins. Regular users won't want or need to see this.

app/views/service_providers/_form.html.erb
```diff
@@ -158,7 +158,7 @@
       and use UL for checkboxes%>
   <fieldset class='usa-fieldset-inputs usa-fieldset'>
     <legend class='usa-sr-only'>Attribute bundle</legend>
-    <%= form.label :attribute_buntle,
+    <%= form.label :attribute_bundle,
```
amoose: mmmmm buntles. 🤤

app/views/service_providers/show.html.erb
jennywagg: Woops! I had done this and accidentally removed the restriction when I put this in a partial 😅

app/helpers/service_provider_helper.rb
```diff
@@ -19,4 +34,33 @@ def sp_logo_preview_path(file_name)
     'https://github.com/18F/identity-idp/blob/master/app/assets/images/sp-logos/' +
       file
   end
+
+  #:reek:FeatureEnvy
+  def yamlized_sp(service_provider)
+    key_from_issuer = JSON.parse(service_provider.to_json).dig('production_issuer').presence ||
+                      service_provider.issuer
+    yamlable_json = { "'#{key_from_issuer}'" => config_hash(service_provider) }
+    yamlable_json.to_yaml.delete('\"')
```
solipet: Noice.

app/helpers/service_provider_helper.rb
```diff
@@ -19,4 +34,33 @@ def sp_logo_preview_path(file_name)
...
+  end
```
solipet: Seeing multiple redirect urls showing up in the yaml like this:
```
redirect_uris: '''[http://localhost:4756/, http://localhost:4756/callbacks,
    http://localhost:4756/logout]'''
```

PR #257, opened 2019-10-14
LG-2109 Edit view in SP layout change

karlaturcios:
Uses key-value view like groups instead of tables, allows flow of long code to horizontally display without going off canvas
localhost_3001_service_providers_4(Laptop with MDPI screen) (1)


karlaturcios:
Not recognizing the Cicleci test results. We removed items out of a table grid view. Could that be cause?

PR #267, opened 2019-12-03
LG-2173: Initial conversion to yarn/webpack

solipet:
Completely moves from using the asset pipeline for JavaScript code to webpack/yarn. 
In particular:
    -  adds `webpacker` gem
    -  adds webpack-dev-server to Procfile
    -  moves existing application javascript to `app/javascripts`
    -  abandons `app/assets/javascripts/selectizer.js` (couldn't tell if this was still used for anything)
    -  removes `report_only` config from secure_headers csp config (this has caused errors to slip through to int)
    -  includes localhost as needed in secure_headers csp config (conditionally in development only)
    -  updates `bin/setup` (this was overdue)


app/javascript/packs/application.js
```diff
@@ -0,0 +1,22 @@
+/* eslint no-console:0 */
+// This file is automatically compiled by Webpack, along with any other files
+// present in this directory. You're encouraged to place your actual application logic in
+// a relevant structure within app/javascript and only use these pack files to reference
+// that code so it'll be compiled.
+
+import 'jquery';
```
jmhooper: Do we use jQuery and UJS in the dashboard? I didn't think we did.

solipet: we do -- the login link has `data-method="post"` on them and without ujs it doesn't POST but GET's instead (which fails)

bin/setup
```diff
@@ -1,29 +1,57 @@
...
+            ___  /____________ ___(_)______   _______ _________   __
+            __  /_  __ \_  __ `/_  /__  __ \  __  __ `/  __ \_ | / /
+            _  / / /_/ /  /_/ /_  / _  / / /___  /_/ // /_/ /_ |/ /
+            /_/  \____/_\__, / /_/  /_/ /_/_(_)\__, / \____/_____/
+                       /____/                 /____/
+        ________              ______ ______                     _________
+        ___  __ \_____ __________  /____  /_____________ _____________  /
+        __  / / /  __ `/_  ___/_  __ \_  __ \  __ \  __ `/_  ___/  __  /
+        _  /_/ // /_/ /_(__  )_  / / /  /_/ / /_/ / /_/ /_  /   / /_/ /
+        /_____/ \__,_/ /____/ /_/ /_//_.___/\____/\__,_/ /_/    \__,_/
```
jennywagg: heh neat

PR #270, opened 2019-12-18
Move "env status" off of home page

blacktm:
Partners are confused by the current "environments status" on the home page, namely suggesting that they can integrate with multiple login.gov environments. This creates unnecessary support issues. The "environments status" page is still useful, so this commit moves it to a new `/env` path.


jmhooper:
@blacktm: Should we link to the new environment page somewhere?

app/controllers/env_controller.rb
```diff
@@ -0,0 +1,12 @@
+# Controller for the "environments status" page
+class EnvController < ApplicationController
```
jmhooper: Looks like this controller has no test coverage. We should probably add a controller spec for it.

config/routes.rb
```diff
@@ -8,6 +8,7 @@
   end
   get '/auth/logindotgov/callback' => 'users/omniauth#callback'
   get 'users/none' => 'users#none'
+  get 'env' => 'env#index'
```
jmhooper: This is a nit, but could we change `get 'env'` to `get '/env'` for consistency.

blacktm: Updated in 692be98

app/controllers/env_controller.rb
blacktm: Added a controller test in 5db53bd (we have the `DeployStatusChecker` test coverage in `/spec/services/deploy_status_checker_spec.rb` already).

spec/controllers/env_controller_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+describe EmailsController do
```
jmhooper: I think you mean `EnvController` here

blacktm: D'oh, fixed, waiting on tests to re-run

blacktm:
@jmhooper Dang it, now getting a `WebMock::NetConnectNotAllowedError`, not sure how to fix. Just needing this "Environments status" page moved so we can stop the support inquiries. Any ideas?

spec/controllers/env_controller_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+describe EnvController do
+  include Devise::Test::ControllerHelpers
+  WebMock.allow_net_connect!
```
zachmargolis: I don't think we should do this .... we should mock the responses we need

spec/features/home.rb
```diff
@@ -0,0 +1,39 @@
+require 'rails_helper'
+
+feature 'Home' do
+  WebMock.allow_net_connect!
```
zachmargolis: This is a bad pattern IMO

spec/features/home.rb
```diff
@@ -0,0 +1,37 @@
...
+    scenario 'you should see intro and links to dev docs and partners' do
+      visit root_path
+
+      expect(page).to have_content(I18n.t('home.body_intro'))
+      expect(page).to have_content(I18n.t('home.dev_docs_link'))
+      expect(page).to have_content(I18n.t('home.partners_link'))
+    end
+  end
+
+  context 'an adminn user' do
```
zachmargolis: nit: typo in `adminn`

spec/support/deploy_status_checker_helper.rb
```diff
@@ -0,0 +1,29 @@
+module DeployStatusCheckerHelper
```
zachmargolis: good idea to extract this! Should we also do the same in `deploy_status_checker_spec.rb`'s `#check` section?

PR #273, opened 2020-01-16
language changes related to teams

karlaturcios:
Could someone take a look at this draft branch? I have a minor rails syntax error on a hint I'm adding to the group label.

### DIFF
--- a/app/views/groups/_form.html.erb
+++ b/app/views/groups/_form.html.erb
@@ -19,7 +19,7 @@
                                  collection_wrapper_tag: :ul,
                                  collection_wrapper_class: 'usa-input-list',
                                  item_wrapper_tag: :li) do |b|,
-                                 hint: 'Select members to add to this team.' %>
+                                 hint: "Select members to add to this team.".html_safe %>
         <%=  b.check_box(class: "usa-checkbox__input", disabled: b.object == current_user && !current_user.admin?) %>
         <%=  b.label(class: "usa-checkbox__label")%>
     <% end %>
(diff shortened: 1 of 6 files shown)


app/views/groups/_form.html.erb
```diff
@@ -18,13 +18,14 @@
                                  :id, :email,
                                  collection_wrapper_tag: :ul,
                                  collection_wrapper_class: 'usa-input-list',
-                                 item_wrapper_tag: :li) do |b| %>
+                                 item_wrapper_tag: :li,
+                                 hint: "Select members to add to this team.".html_safe) do |b| %>
```
jmhooper: I was looking at this and it looks like sneaking it in here doesn't actually display a hint. I think you need to do something like this:

```ruby
<%= form.collection_check_boxes( :user_ids,
                                current_user.admin? ? User.all.sorted : User.select { |u| u.groups.include?(@group) },
                                :id, :email,
                                collection_wrapper_tag: :ul,
                                collection_wrapper_class: 'usa-input-list',
                                item_wrapper_tag: :li) do |b| %>
      <%=  b.check_box(class: "usa-checkbox__input", disabled: b.object == current_user && !current_user.admin?) %>
      <%=  b.label(class: "usa-checkbox__label")%>
  <% end %>
  <%= form.hint "Select members to add to this team." %>
```

jennywagg: Oh good catch, I wasn't careful about what I was moving the hint tag onto

karlaturcios:
Why: The new label for Groups is now Teams so content has been revised in the app 

How:

Update labels in buttons, navigation, forms
Update form hints

karlaturcios:
Thanks for the review @jmhooper

PR #276, opened 2020-01-24
Rename "Group" classes to "Team" (LG-2467)

zachmargolis:
**Why**: Match user-facing naming

Does not rename the actual DB tables and attributes, so there
are a few "group_id" attributes around


app/views/service_providers/_form.html.erb
```diff
@@ -13,11 +13,12 @@
                  label: 'Description',
                  hint: 'A description of the app (may be helpful for your colleagues).' %>
 
-  <%= form.association :group,
+  <%= form.association :team,
```
zachmargolis: this helper was really tricky.... it builds a form key named `service_provider[group_id]` and I couldn't figure out how to override it to be something like `service_provider[team_id]` (I think it uses a lot of reflection to build that), so I had to have the controllers use the old `group_id` attribute

app/controllers/teams_controller.rb
```diff
@@ -0,0 +1,83 @@
+# :reek:InstanceVariableAssumption
+class TeamsController < AuthenticatedController
```
zachmargolis: This is actually the `GroupsController` with a rename....but I guess the changes were big enough that git/Github thinks they are different files

config/routes.rb
```diff
@@ -10,7 +10,8 @@
   get 'users/none' => 'users#none'
 
   resources :users
-  resources :groups
+  resources :groups, controller: 'teams' # preserve old URLs
```
zachmargolis: Do we know if people have bookmarked the `/groups` URL? Do we care? I think there's a case to be made to not keep the old URLs at all

zachmargolis:
The codeclimate errors....all of them existed in the old code. Is that just stricter settings over time? Do we override them? Or should I fix those?

jennywagg:
> The codeclimate errors....all of them existed in the old code. Is that just stricter settings over time? Do we override them? Or should I fix those?

If there's one or two codeclimate complaints that are from existing code, I'd normally just go ahead and fix it - otherwise, if it would take more time (like having twelve errors to address here 😬), then I would normally just merge with the old codeclimate errors. But... I know I personally don't have permissions to do that, so last time I bugged @jmhooper to admin merge for me.

config/routes.rb
jmhooper: Agreed we don't need to keep the old URLs

jmhooper:
> TeamHelper has no descriptive comment

We can we disable that one for the entire codebase?

> Specify an :inverse_of option. 

These looks like ones we may actually want to fix, right? Otherwise ActiveRecord can't figure out the inverse relationship.

zachmargolis:
> Specify an :inverse_of option.

I don't think I've ever relied on `:inverse_of` but I'll add them here

PR #278, opened 2020-01-24
Ruby 2.6

solipet:
Update dashboard to use ruby 2.6.

Included a couple of easy vulnerability updates and deprecation fixes.


Gemfile
```diff
@@ -1,9 +1,10 @@
 source 'https://rubygems.org'
 
-# ruby '~> 2.3.5'
+ruby '~> 2.6'
```
zachmargolis: Should we keep a patch version here like we had been?

Gemfile
```diff
@@ -1,9 +1,10 @@
...
 
 gem 'actionmailer-text', '>= 0.1.1'
 gem 'active_model_serializers', '>= 0.10.7'
+gem 'bundler', '~> 1.17.0'
```
zachmargolis: Talking with @jmhooper, we're gonna see if we can bump this to bundler 2+

PR #280, opened 2020-02-10
Preserves groups route

amoose:
Why
We are utilizing 'dashboard group URL' in HubSpot and these links are now broken.

How
Matches /groups/# URLs and redirects to the Teams route.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -19,4 +19,7 @@ Rails.application.routes.draw do
   resources :service_providers
 
   root to: 'home#index'
+
+  # preserve old Groups route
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -19,4 +19,7 @@
   resources :service_providers
 
   root to: 'home#index'
+
+  # preserve old Groups route
+  match '/groups/:id', to: redirect('/teams/%{id}'), via: :get
```
zachmargolis: Sorry about that, this was me: https://github.com/18F/identity-dashboard/pull/276#discussion_r370419070

amoose: np. this is just easier than fixing all the hubspot links.

PR #281, opened 2020-02-12
[LG-2604] Redirects root route to SP path for logged in users

amoose:
Resolves LG-2604 by redirecting to the service providers path for the root path (when logged in).

### DIFF
--- a/spec/controllers/home_controller_spec.rb
+++ b/spec/controllers/home_controller_spec.rb
@@ -1,0 +1,34 @@
+require 'rails_helper'
+
+describe HomeController do
+  include Devise::Test::ControllerHelpers
+
+  let(:user) { create(:user) }
+  let(:org) { create(:team) }
+
+  before do
+    allow(controller).to receive(:current_user).and_return(user)
+    sign_in user
+  end
+
+  describe '#index' do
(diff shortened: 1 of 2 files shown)


spec/controllers/home_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    context 'when the user is signed in' do
+      it 'has a redirect response' do
+        get :index
+        expect(response.status).to eq(302)
+      end
+    end
+
+    context 'when the user is not signed in' do
+      before do
+        allow(controller).to receive(:current_user).and_return(nil)
```
zachmargolis: I wish there was an easy way to just override the `let(:user)` with `nil` but the `sign_out user` line below is making that tough

spec/controllers/home_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'rails_helper'
+
+describe HomeController do
+  include Devise::Test::ControllerHelpers
+
+  let(:user) { create(:user) }
+  let(:org) { create(:team) }
```
zachmargolis: this looks unused?

PR #282, opened 2020-02-18
LG-2539: Use primary button style for "edit team" button

zachmargolis:
Before

Screen Shot 2020-02-18 at 2 14 11 PM

After

Screen Shot 2020-02-18 at 2 14 02 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,7 @@ jobs:
       - image: postgres:9.6
         environment:
           POSTGRES_USER: circleci
+          POSTGRES_PASSWORD: password
 
     working_directory: ~/identity-dashboard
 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       - image: postgres:9.6
         environment:
           POSTGRES_USER: circleci
+          POSTGRES_PASSWORD: password
```
zachmargolis: This is not strictly related to the PR, but CircleCI was failing due to psql config stuff, so I just figured it would be easier to add in the same PR, happy to move it out if anybody is concerned

karlaturcios: Nice value add! :)

PR #288, opened 2020-03-09
LG-2706 whitelisted users can create/manage teams

jennywagg:
**Why:** Certain users whose emails end in .gov, .mil, or .fed.us are allowed to create teams so they can get started on the dashboard right away.

Also some updates were made (even for non-whitelisted users) to fit these permissions... in particular regarding the deletion of a team:

https://docs.google.com/document/d/17gh18BoeJ84ZZKX_z78xuDGHk4Lko9AfcvtctOqT-y4/edit#

### DIFF
--- a/app/helpers/team_helper.rb
+++ b/app/helpers/team_helper.rb
@@ -7,10 +7,14 @@ module TeamHelper
   end
 
   def can_create_teams?(user)
-    WHITELISTED_DOMAINS.any? { |domain| user.email.end_with? domain } || user.admin?
+    whitelisted_user? || user.admin?
   end
 
   def can_delete_team?(user, team)
     team.users.include?(user) || user.admin?
+  end
+
+  def whitelisted_user?
+    WHITELISTED_DOMAINS.any? { |domain| current_user.email.end_with? domain }
   end
 end
--- a/app/helpers/team_helper.rb
+++ b/app/helpers/team_helper.rb
@@ -1,5 +1,15 @@
 module TeamHelper
+  WHITELISTED_DOMAINS =  %w[.mil .gov .fed.us].freeze
   def can_edit_teams?(user)
     !user.teams.empty? || user.admin?
   end
+
+  def can_create_teams?(user)
+   WHITELISTED_DOMAINS.any? { |domain| user.email.end_with? domain } || user.admin?
+  end
+
+  def can_delete_team?(user, team)
+   team.users.include?(user) || user.admin?
+  end
+
 end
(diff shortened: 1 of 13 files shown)


config/initializers/simple_form.rb
```diff
@@ -45,7 +45,7 @@
     # changed input class from field to usa-input
     b.use :input, class: 'block col-12' # usa-input'
     b.use :hint,  wrap_with: { tag: 'div', class: 'usa-form-hint' }
-    b.use :error, wrap_with: { tag: 'div', class: 'usa-error-message margin-top-neg-205' }
+    b.use :error, wrap_with: { tag: 'div', class: 'usa-error-message' }
```
jennywagg: @karlaturcios I know you added a class here to solve some error message alignment issues, but on the teams page it actually causes an error message to be out of alignment, so I removed it. If you could show me which different error message on another form this fixed maybe we can figure out what's going on 🤔

app/policies/team_policy.rb
```diff
@@ -23,15 +25,15 @@ def edit?
   end
 
   def destroy?
-    admin?
+    in_team? || admin?
```
jennywagg: Wanted to acknowledge that this seems scary but it follows this:

https://docs.google.com/document/d/17gh18BoeJ84ZZKX_z78xuDGHk4Lko9AfcvtctOqT-y4/edit#

solipet: yeah, I made a comment in that doc... if a team is deleted, do we delete all of the apps associated with that team? And are teams really deleted all that often that we need to make it easy for partners?

app/helpers/team_helper.rb
```diff
@@ -1,5 +1,16 @@
+#:reek:UtilityFunction
 module TeamHelper
+  WHITELISTED_DOMAINS = %w[.mil .gov .fed.us].freeze
+
   def can_edit_teams?(user)
     !user.teams.empty? || user.admin?
   end
+
+  def can_create_teams?(user)
+    WHITELISTED_DOMAINS.any? { |domain| user.email.end_with? domain } || user.admin?
```
solipet: ~~Maybe consider just returning true if admin, otherwise do the domain check? As is, the admin check gets lost at the end of a long, busy line.~~

Actually, you could move the `whitelisted_user?` method from the policy to this helper, and include this helper in the policy.

app/policies/team_policy.rb
solipet: I agree!

PR #289, opened 2020-03-10
LG-2751: upgrade dashboard to Rails 5.2

solipet:
Why? Well, aside from the benefits of staying up-to-date, we want to use ActiveStorage to allow partners to upload logos and that feature is only available in Rails 5.2+.

This also upgrades the `simple_form` gem to take care of a long outstanding security vulnerability. :-) 

This is for a subtask of LG-1386.


app/views/service_providers/_form.html.erb
```diff
@@ -136,9 +136,17 @@
                      class: 'usa-label' %>
     </div>
 
-    <%= form.input :redirect_uris,
-                   input_html: { class: 'usa-input', multiple: 'multiple' },
-                   label: false %>
+    <% values = service_provider.redirect_uris || [] %>
+    <% values << nil %>
```
zachmargolis: is there a controller we can move this logic to? IMO it's better/more understandable to have the templates be simple view logic (ifs, loops, etc)

bin/update
```diff
@@ -0,0 +1,31 @@
...
+chdir APP_ROOT do
+  # This script is a way to update your development environment automatically.
+  # Add necessary update steps to this file.
+
+  puts '== Installing dependencies =='
+  system! 'gem install bundler --conservative'
+  system('bundle check') || system!('bundle install')
+
+  # Install JavaScript dependencies if using Yarn
+  # system('bin/yarn')
```
zachmargolis: Does this repo use yarn?

config/environments/development.rb.bak
```diff
@@ -0,0 +1,29 @@
+Rails.application.configure do
```
zachmargolis: Are we going to remove this file?

config/environments/test.rb.bak
```diff
@@ -0,0 +1,17 @@
+Rails.application.configure do
```
zachmargolis: Ditto, we're removing this file right?

config/initializers/application_controller_renderer.rb
```diff
@@ -0,0 +1,8 @@
+# Be sure to restart your server when you modify this file.
```
zachmargolis: Is this a Rails 5.2 default? Are we keeping it?

spec/factories/service_providers.rb
```diff
@@ -18,5 +18,13 @@
     trait :team do
       association :team, factory: :team
     end
+
+    trait :with_users_team do
+      after(:build) do |service_provider|
+        team = service_provider.user&.teams[1] || create(:team)
```
zachmargolis: I'm not sure I follow this line, is the goal to create a second team if one does not exist already?

config/initializers/application_controller_renderer.rb
zachmargolis: And same for all the other initializers that are commented out

bin/update
solipet: Yup!

config/environments/development.rb.bak
solipet: doh!

config/environments/test.rb.bak
solipet: ditto, doh!

config/initializers/application_controller_renderer.rb
solipet: I've chosen to revert all of the config files generated by rails back to what they were, with only the specific changes we make to them applied. For example, I'm not deleting all of the comments that come with them. This makes it so much easier to run diffs when upgrading to a new version. As it was, I had to pick through every line in our configs to map them to the 5.2 versions.

app/views/service_providers/_form.html.erb
solipet: Yup - good call

spec/factories/service_providers.rb
solipet: Good point -- if the user doesn't have a team, the team will be nil and the test will (presumably) fail, which likely indicates a problem with the test set up.

spec/factories/service_providers.rb
```diff
@@ -18,5 +18,13 @@
     trait :team do
       association :team, factory: :team
     end
+
+    trait :with_users_team do
+      after(:build) do |service_provider|
+        team = service_provider.user&.teams[0]
```
zachmargolis: Sorry the goal of this still isn't clear to me, is it just making sure that the SP has a team if it has a user? What is the goal? And it's optional, right?

app/helpers/service_provider_helper.rb
```diff
@@ -35,6 +35,12 @@ def sp_logo_preview_path(file_name)
       file
   end
 
+  # Generate the list for the SP edit form, including a nil entry
+  def redirect_uri_list(service_provider = @service_provider)
```
zachmargolis: Since we only call this from one place, can we drop the default argument? Either just use `@service_provider` in the helper or put `@service_provider` directly in the template as an argument?

My thinking is that it makes future refactoring harder if you have to guess

PR #293, opened 2020-03-12
LG-2762 move delete team button

jennywagg:
**Why:**  Because the delete button doesn't belong on the teams list cards

### DIFF
--- a/app/assets/stylesheets/application.css.scss.erb
+++ b/app/assets/stylesheets/application.css.scss.erb
@@ -10,7 +10,6 @@ $image-path: 'img';
 @mixin text-wrap-mixin {
   word-break: break-all;
 }
-
 .text-wrap {
   @include text-wrap-mixin;
 }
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/application.css.scss.erb
```diff
@@ -12,4 +12,14 @@ $image-path: 'img';
 }
 .text-wrap {
   @include text-wrap-mixin;
+}
+
+.button-delete {
+  color: white !important;
+  text-decoration: none !important;
```
jennywagg: This is bad. I've discussed with @karlaturcios the need for a "link-button" (or "button-link"?) style in the design system, since using links that behave as buttons is a valid solution in a case like this, where we have multiple buttons trying to perform different actions in/around a form.

karlaturcios: We will create a new DS component for this and come back to remove local styles. Users need the delete button on this screen ASAP.

PR #294, opened 2020-03-13
LG-1386: Partner self-service logo upload

solipet:
As a login.gov partner, integrating with login.gov,
I want to be able to upload my agency's logo file, if not already available,
so that I can see the co-branded page on Login.gov during integration.

Still needed:
* Feature flag (off by default)
* Documentation update
* integration tests?
* design refinements?
** currently using custom elements/css -- should create a new story to make a component in the style guide


spec/fixtures/invalid.txt
```diff
@@ -0,0 +1 @@
+invalid
```
jennywagg: Can someone help me understand what this is 🤔

jmhooper: This is a test fixture. It's basically a file that we use in our tests. This one is used to test that the logo upload rejects files that are not PNGs or SVGs.

PR #297, opened 2020-03-17
LG-2471 Update authenticated user homepage

jennywagg:


app/views/home/authenticated/index.erb
```diff
@@ -0,0 +1,20 @@
+<div class="usa-alert usa-alert--info">
+  <div class="usa-alert__body">
+    <p class="usa-alert__text">You can now create teams.</p>
```
jmhooper: So, can't only government users create teams? Unless I'm mistaken it looks like everyone will see this message telling them they can.

jennywagg: Thanks, I foolishly missed this 😬

PR #300, opened 2020-03-18
Add logo file info to the service provider API response

jmhooper:
**Why**: So the IdP can use this to get the logos from S3.


jmhooper:
Chatted with Amit and we're actually going to approach this a little differently

PR #308, opened 2020-03-24
LG-2813 Improve admin landing page, and other improvements

jennywagg:
**Why?**  We want admins to have an easy-to-read landing page on the dashboard that only includes the teams of which they are members, rather than all Partnership teams on the landing page.

Snuck some other things into this PR:

- Make separate pages for an admin user's teams and an admin-only view of "All teams" ([LG-2815](https://cm-jira.usa.gov/browse/LG-2815))
       - Restricted access for 'all' pages to admins only (like the new `/teams/all`, and `/service_providers/all`, which was accessible by any partner before)

- Updated buttons for apps on team cards ([LG-2814](https://cm-jira.usa.gov/browse/LG-2814))
- Allow "0" label to show when there are no apps on a team card ([LG-2817](https://cm-jira.usa.gov/browse/LG-2817))
- Quick fix to admin SP yaml blurb ('cert' key is correct) ([LG-2696](https://cm-jira.usa.gov/browse/LG-2696))

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -3,11 +3,7 @@ class HomeController < ApplicationController
     render :index and return unless user_signed_in?
 
     includes = %i[users service_providers agency]
-    @teams = if current_user.admin?
-               Team.includes(*includes).all
-             else
-               current_user.teams.includes(*includes).all
-             end
+    @teams = current_user.teams.includes(*includes).all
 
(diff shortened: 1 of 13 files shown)


karlaturcios:
@jennywagg is it possible to have a separate  header view for admin vs authenticated like the mockup?
admin with teams

jennywagg:
image

@karlaturcios Updated it!

PR #312, opened 2020-04-06
LG-2714: new rake task to import legacy logos into active storage

solipet:
To prepare for the conversion to partner uploaded logos, we need to attach all of the existing legacy logos to the current SP's.

`rake service_providers:import_legacy_logos` will clone (or update) `identity-idp-config` and then for each sp returned from `/api/service_providers`, find the logo file in the config repo that is named by the sp and attach it to the sp via ActiveStorage.

On a system that is using S3 for ActiveStorage, this should result in the image being uploaded.


solipet:
The last codeclimate complaint is that the main method is too complex. Personally, I disagree, and in any case, this is code that will likely be run one time then thrown away so I think we should merge it without worrying about the perceived complexity.

Gemfile
```diff
@@ -12,6 +12,7 @@ gem 'dotenv-rails', '~> 2.4'
 gem 'enum_help'
 gem 'figaro'
 gem 'httparty'
+gem 'identity-hostdata', github: '18F/identity-hostdata', branch: 'master'
```
zachmargolis, line 15: 😍

app/helpers/service_provider_helper.rb
```diff
@@ -1,4 +1,4 @@
-#:reek:UtilityFunction
+#:reek:UtilityFunction, :reek:TooManyConstants
```
zachmargolis, line 1: note to self, time to just get rid of reek

app/services/service_provider_logo_updater.rb
```diff
@@ -0,0 +1,123 @@
...
+  # :reek:TooManyStatements
+  # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+  def import_logos_to_active_storage
+    return unless Figaro.env.logo_upload_enabled
+
+    git_latest_idp_config
+    idp_config.each do |sp|
+      issuer = sp['issuer']
+      logo_name = sp['logo']
+      logger.info('~~ ' + issuer.to_s + ' logo: ' + logo_name.to_s)
```
zachmargolis, line 16: are we leaving this in intentionally?

app/services/service_provider_logo_updater.rb
```diff
@@ -0,0 +1,123 @@
...
+      'images',
+      'sp-logos',
+      filename
+    )
+  end
+
+  #############################################################################
+
+  def update_repo
+    cmd = ['cd', repo_dir, ';', 'git', 'pull']
```
zachmargolis, line 68: I don't like `cd` without changing back...what if we did:

```
Dir.chdir(repo_dir) do
  # git pull ... etc
end
```

zachmargolis: Unsure if subprocess resets the working dir after commands, but I think this form is clearer

app/services/service_provider_logo_updater.rb
```diff
@@ -0,0 +1,123 @@
...
+    logger = Logger.new(STDOUT)
+    logger.progname = 'import_logos_to_active_storage'
+    logger
+  end
+
+  #############################################################################
+
+  # :reek:UtilityFunction
+  def idp_config_checkout_name
+    'identity-idp-config'.freeze
```
zachmargolis, line 121: why not just make this a constant?

app/services/service_provider_logo_updater.rb
```diff
@@ -0,0 +1,123 @@
...
+  #############################################################################
+
+  def update_repo
+    cmd = ['cd', repo_dir, ';', 'git', 'pull']
+    logger.info(cmd.join(' '))
+    Subprocess.check_call(cmd)
+  end
+
+  def clone_repo
+    repo_url = ENV.fetch('IDP_private_config_repo',
```
zachmargolis, line 74: do we usually mix casing of variables like this?

app/services/service_provider_logo_updater.rb
```diff
@@ -0,0 +1,123 @@
...
+      clone_repo
+    end
+  end
+
+  def idp_config
+    @idp_config ||= load_idp_config
+  end
+
+  def handle_error(error)
+    # ::NewRelic::Agent.notice_error(error)
```
zachmargolis, line 49: do we want to keep this?

zachmargolis:
@solipet 

> The last codeclimate complaint is that the main method is too complex. Personally, I disagree, and in any case, this is code that will likely be run one time then thrown away so I think we should merge it without worrying about the perceived complexity.

Maybe just update the `.reek` to ignore the file completely?

app/services/service_provider_logo_updater.rb
solipet: I think it's helpful for the person running the rake task to see that it's doing something. Could make some output that is less obnoxious?

zachmargolis: Yup, that's useful output. I use uncommon characters like `~` and `====` when doing test prints locally so I wanted to make sure this was left intentionally

app/services/service_provider_logo_updater.rb
solipet: no, I copied this from the idp's activate.rb :-|  I'll change it here.

app/services/service_provider_logo_updater.rb
solipet: nope

app/services/service_provider_logo_updater.rb
solipet: Yup

app/services/service_provider_logo_updater.rb
solipet: Good call.

app/helpers/service_provider_helper.rb
solipet: 😍

spec/services/service_provider_logo_updater_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    )
+  end
+
+  describe '#import_logos_to_active_storage' do
+    it 'attaches the legacy logo to the service provider' do
+      # don't clone the config repo
+      allow(Subprocess).to receive(:check_call).and_return(true)
+      # don't call /api/service_providers on the dashboard
+      # allow(updater).to receive(:load_idp_config).and_return(fake_config)
+      allow_any_instance_of(RestClient::Request).to receive(:execute).and_return(fake_response)
```
zachmargolis, line 22: we should be able to stub this with webmock's `stub_request` instead of having to do `allow_any_instance_of` (ditto for the rest of this file too)

spec/services/service_provider_logo_updater_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      body: JSON.generate(fake_config)
+    )
+  end
+
+  describe '#import_logos_to_active_storage' do
+    it 'attaches the legacy logo to the service provider' do
+      # don't clone the config repo
+      allow(Subprocess).to receive(:check_call).and_return(true)
+      # don't call /api/service_providers on the dashboard
+      # allow(updater).to receive(:load_idp_config).and_return(fake_config)
```
zachmargolis, line 21: remove this commented out line? ditto for the other spec too?

PR #314, opened 2020-04-06
Remove reek

zachmargolis:
See https://github.com/18f/identity-idp/pull/3547


.codeclimate.yml
```diff
@@ -59,16 +59,6 @@ plugins:
       - XXX
   pep8:
     enabled: true
-  reek:
-    enabled: true
-    exclude_patterns:
-    - 'spec/**/*'
-    - 'db/**/*'
-  rubocop:
```
jennywagg, line 67: In reference to my other comment - to be clear the idea is we're just getting rid of it in codeclimate, right? Keep it around in the app though?

zachmargolis: whoooops no I just meant to remove reek and keep rubocop 👮

jennywagg: Oh hehe

zachmargolis: Thanks for catching!

PR #316, opened 2020-04-09
lg-2794 adding toggle to indicate testing or prod for new SPs

shellzie:
Why? so a SP can specify test or prod

### DIFF
--- a/app/javascript/app/service-provider-form.js
+++ b/app/javascript/app/service-provider-form.js
@@ -75,6 +75,9 @@ $(function(){
       logo_file = input.files[0];
       preview.textContent = logo_file.name;
     });
-    $('input[id="service_provider_environment_0"]').attr('checked', true);
   }
+
+  $(document).ready(function() {
+    $('input[id="service_provider_environment_int"]').attr('checked', true);
+  });
 });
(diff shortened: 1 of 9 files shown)


app/models/service_provider.rb
```diff
@@ -17,6 +17,7 @@ class ServiceProvider < ApplicationRecord
 
   enum block_encryption: { 'none' => 0, 'aes256-cbc' => 1 }, _suffix: 'encryption'
   enum identity_protocol: { openid_connect: 0, saml: 1 }
+  enum environment: { 'Testing': 0, 'To go live in production': 1 }
```
zachmargolis, line 20: what if we named the enums for the environments we have?
- `int`
- `staging`
- `prod`
- `dev`

shellzie:
updated the enum to use strings instead of ints. will go back and add a test if this is approved.

app/javascript/app/service-provider-form.js
```diff
@@ -76,4 +76,8 @@ $(function(){
       preview.textContent = logo_file.name;
     });
   }
+
+  $(document).ready(function() {
```
jmhooper, line 80: Looks like the JS here is always checking this. Is there a reason to do that with JS instead of when we render the form?

PR #319, opened 2020-04-13
Upgrade Pundit to v2

jmhooper:
**Why**: Because v2 of the gem has a `policy_class` option on `#authorize`. This is going to be helpful for me when I'm working on bulk adding users and authorizing that action.


jmhooper:
Total coverage drop? Seriously?

jmhooper:
Trying a rebase against master to see if that will fix the code coverage drop

Gemfile.lock
```diff
@@ -290,7 +290,7 @@ GEM
     public_suffix (4.0.3)
     puma (4.3.3)
       nio4r (~> 2.0)
-    pundit (1.1.0)
+    pundit (2.1.0)
```
zachmargolis, line 293: if we want >2 for a feature should we be explicit about that in `Gemfile` too?

jmhooper: Doesn't hurt

PR #320, opened 2020-04-13
Add the webdrivers gem

jmhooper:
**Why**: So that the webdrivers gem can figure out what version of chromedriver to use instead of me


jmhooper:
Also rebasing here to see if that fixes code coverage drop

PR #323, opened 2020-04-14
LG-2882 prompt user to delete apps before deleting non-empty team

jennywagg:
**Why:** Because we don't want to delete teams which still have apps attached to them, therefore orphaning those apps.

### DIFF
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -37,11 +37,11 @@ class TeamsController < AuthenticatedController
   def destroy
     if @team.service_providers.empty? && @team.destroy
       flash[:success] = 'Success'
-      redirect_to teams_path and return
+      return redirect_to teams_path
     end
 
     flash[:warning] = I18n.t('notices.team_delete_failed')
-    redirect_back(fallback_location: teams_path) and return
+    redirect_back(fallback_location: teams_path)
   end
 
   def index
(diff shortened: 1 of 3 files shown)


app/controllers/teams_controller.rb
```diff
@@ -35,9 +35,13 @@ def update
...
-    return unless @team.destroy
-    flash[:success] = 'Success'
-    redirect_to teams_path
+    if @team.service_providers.empty? && @team.destroy
+      flash[:success] = 'Success'
+      redirect_to teams_path and return
+    end
+
+    flash[:warning] = I18n.t('notices.team_delete_failed')
+    redirect_back(fallback_location: teams_path) and return
```
jmhooper, line 44: Do you need the `and return` on this one since it is the last line in the method?

app/controllers/teams_controller.rb
```diff
@@ -35,9 +35,13 @@ def update
   end
 
   def destroy
-    return unless @team.destroy
-    flash[:success] = 'Success'
-    redirect_to teams_path
+    if @team.service_providers.empty? && @team.destroy
+      flash[:success] = 'Success'
+      redirect_to teams_path and return
```
jmhooper, line 40: This may be on the personal preference side so feel free to take or leave. I find `return redirect_to(team_path)` a little more readable.

app/controllers/teams_controller.rb
jennywagg: probably not!

PR #324, opened 2020-04-15
LG-2715: publish logo key

solipet:
When a service provider has a logo attached, that logo is stored (either locally or in S3) in a file with a filename that can be retrieved by calling `logo_file.key`. This PR publishes that key to the JSON endpoint `/api/service_providers` so that it may be received by the IdP and saved with its copy of the SP and then used to access the logo image directly off of S3.

Also cleans up the attaching of logo files including setting the filename in the legacy field `logo`, forcing the content-type to be set on the S3.


config/environments/development.rb
```diff
@@ -30,6 +30,7 @@
 
   # Store uploaded files on the local file system (see config/storage.yml for options)
   config.active_storage.service = :local
+  config.active_storage.content_types_to_serve_as_binary -= ['image/svg+xml']
```
zachmargolis, line 33: maybe drop a comment about how we are explicitly trusting this because are going to be referencing these from `` tags which don't execute scripts?

app/controllers/service_providers_controller.rb
```diff
@@ -135,6 +134,37 @@ def logo_file_param
...
+  def attach_logo_file
+    return unless logo_file_param
+
+    service_provider.logo_file.attach(logo_file_param)
+    service_provider.logo = service_provider.logo_file.filename.to_s
+    push_logo_content_type
+  end
+
+  # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+  def push_logo_content_type
```
zachmargolis, line 146: can we add some sort of coverage of this? like a simple unit test that makes sure we call to s3?

jmhooper: I'm looking at this and thinking it may be easier to test if we move it into a service

jmhooper: Another thing that may be helpful is client stubbing which you get from the AWS SDK: https://aws.amazon.com/blogs/developer/advanced-client-stubbing-in-the-aws-sdk-for-ruby-version-3/

solipet: Using the client stubbing made the tests much easier :-) I'm going to hold off on refactoring to a service for now.

PR #325, opened 2020-04-16
Move user management functionality to a new screen / controller

jmhooper:
**Why**: So we can allow for better adding / removing of users away from the Edit Team page.


app/forms/manage_users_form.rb
```diff
@@ -0,0 +1,37 @@
...
+    result = team.update(users: users_from_user_emails)
+    errors.merge!(team.errors)
+    result
+  end
+
+  private
+
+  def users_from_user_emails
+    user_emails.map do |email|
+      User.find_by(email: email).presence || User.new(email: email)
```
zachmargolis, line 27: Do you think it would be worth trying to combine this into one DB query instead of many?

```ruby
existing_users = User.where(email: user_emails).to_a
missing_users = (user_emails - existing_users.map(&:emails)).map { |e| User.new(email: e) }

(existing_users + missing_users).sort_by(&:email)
```

app/javascript/app/manage_users/email_form_input.js
```diff
@@ -0,0 +1,8 @@
+export const buildEmailAddressFormInput = (email) => {
+  const input = document.createElement("input");
+  input.type = "hidden";
+  input.name = "user_emails[]";
+  input.className = "user_email_input";
+  input.value = email;
+  return input;
```
zachmargolis, line 7: IMO this is starting to veer into "should we just use a JS templating library" territory. Combined with the `updateEmailAddressList` in the other file, I feel like this would make for a great mustache template. I know mustache doesn't support callbacks out of the box.... so maybe that's a case for a small React form or Vue.js (which I haven't used) seems to support event handlers as well.

jmhooper: Yeah, I thought several times about pulling in React to handle this. It felt a little heavy handed. If you feel strongly I can go down that road though.

app/forms/manage_users_form.rb
jmhooper: Nice, I like it

app/javascript/app/manage_users/email_form_input.js
zachmargolis: You're the author, and work in this codebase more than I do at the moment, so I'll defer to you! Just saying that I think we're definitely approaching the point where it makes sense to switch to something soon.

jmhooper:
@solipet:

The hitting enter there is definitely annoying. Let me see what I can do about that.

For the validation, how would you feel about coming back and picking that one up in another story?

solipet:
> @solipet:
> 
> The hitting enter there is definitely annoying. Let me see what I can do about that.
> 
> For the validation, how would you feel about coming back and picking that one up in another story?

sounds good

jmhooper:
@solipet: I cleaned up the enter button behavior in this commit: https://github.com/18F/identity-dashboard/pull/325/commits/f58cf104d06e9b36b17ccc36f793c41c29bce882

PR #328, opened 2020-04-21
LG-2871: have logo uploads enabled by default

solipet:
Given the difficulty in sorting out how to actually push out changes to figaro settings without a devops PR, we're going to simply turn on logo uploads by default. They may still be turned off via figaro, and the bucket is still configurable via figaro.

If I can get this merged asap, I'll test out in `dev` prior to it getting auto deployed to `int`.

### DIFF
--- a/app/views/service_providers/show.html.erb
+++ b/app/views/service_providers/show.html.erb
@@ -29,7 +29,7 @@
       <p class="font-mono-xs margin-top-0" name="production_issuer"><%=  service_provider.production_issuer %></p>
     <% end %>
 
-<% if Figaro.env.logo_upload_enabled == 'true' %>
+<% if ENV.fetch('logo_upload_enabled') { 'true' } == 'true' %>
   <h4><label for="logo_file">Uploaded Logo:</label></h4>
   <% if service_provider.logo_file.attached? %>
     <p class="font-mono-xs margin-top-0" name="logo_file">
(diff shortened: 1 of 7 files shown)


app/views/service_providers/show.html.erb
```diff
@@ -29,7 +29,7 @@
       <p class="font-mono-xs margin-top-0" name="production_issuer"><%=  service_provider.production_issuer %></p>
     <% end %>
 
-<% if Figaro.env.logo_upload_enabled == 'true' %>
+<% if ENV.fetch('logo_upload_enabled') { 'true' } == 'true' %>
```
zachmargolis, line 32: I'd just do the direct form not the block form: `ENV.fetch('logo_upload_enabled', 'true') == 'true')`

config/application.yml.example
```diff
@@ -1,10 +1,10 @@
 test: &default
   aws_region: us-west-2
-  aws_logo_bucket: 'changeme'
+  # aws_logo_bucket: 'changeme'
```
zachmargolis, line 3: should we leave a note as to why it's commented out?

config/environments/development.rb
```diff
@@ -29,7 +29,12 @@
   end
 
   # Store uploaded files on the local file system (see config/storage.yml for options)
-  config.active_storage.service = :local
+  logo_enabled = ENV.fetch('logo_upload_enabled') { 'true' } == 'true'
+  config.active_storage.service = if logo_enabled
```
zachmargolis, line 33: does a ternary fit here?

```ruby
  config.active_storage.service = logo_enabled ? :amazon : :local
```

app/views/service_providers/show.html.erb
zachmargolis: ditto for the other spots too

PR #334, opened 2020-05-01
LG-2954: stop logging to STDOUT in production

solipet:
The current infrastructure design has the standard log file being streamed to AWS CloudWatch. Early on in development, the dashboard was outfitted with the `rails_12factor` gem that, among other things, pulls in the `rails_stdout_logging` gem and forces the Rails logger to write to STDOUT. As such, we've had no access to the deployed dashboard logs since at least March 2018.

For now, we should back out the `rails_stdout_logger` gem and log to the default log file.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,7 +85,7 @@ end
 
 group :production do
   gem 'rack-timeout'
-  gem 'rails_12factor'
+  gem 'rails_serve_static_assets'
 end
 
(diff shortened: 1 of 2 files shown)


zachmargolis:
In a future PR, maybe worth adding in `lograge` to here too and do JSON-formatted logging to be consistent with the IDP?

PR #335, opened 2020-05-04
LG-2871: ready for logo uploads to go live

solipet:
PR #329 reverted the default enabling of logo uploads for partners. That was necessary because we didn't have a convenient way to turn off the feature, and the uploads were blocked by our body size limit of 10k (most logo files are bigger than 10k). The upload limit has been increased to 2MB, so we should be able to run with this now.

### DIFF
--- a/app/views/service_providers/_form.html.erb
+++ b/app/views/service_providers/_form.html.erb
@@ -75,7 +75,7 @@
                    input_html: { class: 'usa-input' } %>
   <% end %>
 
-  <% if Figaro.env.logo_upload_enabled == 'true' %>
+  <% if ENV.fetch('logo_upload_enabled', 'true') == 'true' %>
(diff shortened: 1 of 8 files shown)


app/views/service_providers/_form.html.erb
```diff
@@ -75,7 +75,7 @@
                    input_html: { class: 'usa-input' } %>
   <% end %>
 
-  <% if Figaro.env.logo_upload_enabled == 'true' %>
+  <% if ENV.fetch('logo_upload_enabled', 'true') == 'true' %>
```
zachmargolis, line 78: What's with the switch from Figaro to raw environment variables?

Don't we use Figaro populate environment variables from the YML files? Like my expectation is that if we left it as Figaro, it would accept env vars too, so that's easier to work with?

PR #337, opened 2020-05-08
Update validations to include custom scheme

achapm:


Gemfile.lock
```diff
@@ -20,10 +20,11 @@ GIT
 
 GIT
   remote: https://github.com/18f/identity-validations.git
-  revision: 31d041f479d5479940b9c6c67a3ee37786ace68c
+  revision: 579e79345e6af31013df19f6316e9aebf2686dd4
+  579e79345e6af31013df19f6316e9aebf2686dd4
```
zachmargolis, line 24: this looks like a syntax error to me....? also don't we need to wait for the other PR to merge before we can get the new SHA on the `master` branch?

achapm: updated

PR #338, opened 2020-05-13
LG-2960: set region on s3 client

solipet:


app/controllers/service_providers_controller.rb
```diff
@@ -166,7 +166,7 @@ def using_s3?
   end
 
   def s3
-    @s3 ||= Aws::S3::Client.new
+    @s3 ||= Aws::S3::Client.new region: Figaro.env.aws_region
```
zachmargolis, line 169: style: I'd use parens here please but no big deal

app/services/service_provider_logo_updater.rb
```diff
@@ -105,6 +105,8 @@ def dashboard_sp_url
     else
       'http://localhost:3001/api/service_providers'
     end
+  rescue LoginGov::Hostdata::MissingConfigError
```
zachmargolis, line 108: Looking at the source of the gem....why is this rescue needed? the `in_datacenter?` guard should cover this? Are we stubbing part of the gem or something in specs? If so, I'd look into how we can maybe not stub it, or maybe use `Hostdata.reset!` to clean up any garbage state?

solipet: yup, too much stubbing

solipet:
seriously? diff coverage?!

zachmargolis:
(Used admin privs to merge and override the code coverage diff errors, they were not significant)

PR #340, opened 2020-05-20
LG-2960: clean up logo upload

solipet:
I had disabled logo upload in `int` - that is reversed.
Went back to `Figaro.env` calls vs. `ENV.fetch...`
Configured `development` to use local storage (logos won't be available to idp)
Configured active storage to make service urls expire in 1 week (was 5 min :-o)

### DIFF
--- a/config/initializers/active_storage.rb
+++ b/config/initializers/active_storage.rb
@@ -1,0 +1,1 @@
+ActiveStorage::Service.url_expires_in = 1.week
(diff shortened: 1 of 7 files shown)


config/initializers/active_storage.rb
```diff
@@ -0,0 +1 @@
+ActiveStorage::Service.url_expires_in = 1.week
```
zachmargolis, line 1: Is this something we can set from the YAML config somehow?

solipet: 1 week is the max - I'd prefer permanent. If we're still experiencing the bug Im seeing now after this, we'll have to do a hack where is the storage service is amazon, we provide the absolute url to the public image on S3/Cloudfront instead of the typical `url_for(logo_file)`.

PR #342, opened 2020-05-21
LG-2971: auto-create account for .gov or .mil users

solipet:
As a partner with a `.gov` or `.mil` email address, I don't want to have to email Tom to get access to the login.gov dashboard so I can get started right away. 

Whitelisted TLDs are configurable via `application.yml`. Note, they actually don't have to be TLDs, just the end of the email, e.g., `sa.gov` would whitelist `gsa.gov` and `nasa.gov`.

### DIFF
--- a/app/controllers/users/omniauth_controller.rb
+++ b/app/controllers/users/omniauth_controller.rb
@@ -1,8 +1,12 @@
 module Users
   class OmniauthController < ApplicationController
+    # rubocop:disable Metrics/MethodLength
     def callback
       omniauth_info = request.env['omniauth.auth']['info']
-      @user = User.find_by(email: omniauth_info['email'])
+      email = omniauth_info['email']
+      @user = User.find_by(email: email)
+      allow_unregistered_government_user(email)
+
       if @user
         @user.update!(uuid: omniauth_info['uuid'])
         sign_in @user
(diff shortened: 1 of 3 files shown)


app/controllers/users/omniauth_controller.rb
```diff
@@ -13,5 +17,16 @@ def callback
...
       end
     end
+    # rubocop:enable Metrics/MethodLength
+
+    private
+
+    def allow_unregistered_government_user(received_email)
+      return if @user
+      allowed_tlds = Figaro.env.auto_account_creation_tlds.split(',')
+      return unless allowed_tlds.include?(received_email[-4..-1])
```
zachmargolis, line 27: personally I don't like numerically indexing a string like this.... what about a `.split('.').last` to get the TLD?

zachmargolis: And then maybe have the config be like `gov,mil` instead of `.gov,.mil` ?

solipet: There was talk of allowing `.fed.us` so `.split('.').last` wouldn't work, but then, neither would `[-4..-1]`!  I'll revisit...

solipet: @zachmargolis please take another look. This should work for all cases.

app/controllers/users/omniauth_controller.rb
```diff
@@ -24,7 +24,9 @@ def callback
     def allow_unregistered_government_user(received_email)
       return if @user
       allowed_tlds = Figaro.env.auto_account_creation_tlds.split(',')
-      return unless allowed_tlds.include?(received_email[-4..-1])
+      return if allowed_tlds.filter do |tld|
+        /(#{Regexp.escape(tld)})$/.match?(received_email)
```
zachmargolis, line 28: User `\Z` as an anchor? it's more complete than `$` because it matches end of input and not end of line (in case somebody sneaks in a multi-line email address)

zachmargolis: Also maybe instead of `.filter {}.empty?` use `.none?`

```ruby
return if allowed_tlds.none? |tld|
  # etc
end
```

PR #343, opened 2020-05-21
LG-2791: doesn't specify a whitelist of allowed tld's in deployed envs

solipet:
Removes the tld whitelist from the default so that we can more precisely control each env via app.yml.

### DIFF
--- a/app/controllers/users/omniauth_controller.rb
+++ b/app/controllers/users/omniauth_controller.rb
@@ -23,7 +23,7 @@ module Users
 
     def allow_unregistered_government_user(received_email)
       return if @user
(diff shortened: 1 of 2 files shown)


app/controllers/users/omniauth_controller.rb
```diff
@@ -23,7 +23,7 @@ def callback
 
     def allow_unregistered_government_user(received_email)
       return if @user
-      allowed_tlds = Figaro.env.auto_account_creation_tlds.split(',')
+      allowed_tlds = Figaro.env.auto_account_creation_tlds&.split(',') || []
```
zachmargolis, line 26: Another option is:

```ruby
(Figaro.env.auto_account_creation_tlds || '').split(',')
```

but otherwise same same

PR #345, opened 2020-05-27
LG-2974: enable PaperTrail versioning of ServiceProviders

solipet:
As a user, I would like anything uploaded to the dashboard and promoted to the IdP to include an audit trail, so that if something malicious is uploaded the login team can do forensics to prevent it in the future.

Adds dependency on PaperTrail to version ServiceProviders.

Migrations to add tables required by PT, and to add a cache of `remote_logo_key`, the ActiveStorage key, to ServiceProvider that would not otherwise not be recorded in the object changes.


app/controllers/application_controller.rb
```diff
@@ -1,6 +1,8 @@
 class ApplicationController < ActionController::Base
   include Pundit
 
+  before_action :set_paper_trail_whodunnit
```
zachmargolis, line 4: wow omg that is the most casual name I've seen in a library

PR #347, opened 2020-05-27
LG-1565 add return url

slj:
**Why**: To redirect the user to their requested URL after login
**How**: Save request in session and redirect after authentication

Example: The user tries to visit '/teams' before authenticating.  After logging in, the user is redirected to '/teams' instead of the home page.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,7 @@ class ApplicationController < ActionController::Base
   include Pundit
 
   before_action :set_paper_trail_whodunnit
+  before_action :set_requsted_url
 
   protect_from_forgery with: :exception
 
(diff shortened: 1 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -2,6 +2,7 @@ class ApplicationController < ActionController::Base
   include Pundit
 
   before_action :set_paper_trail_whodunnit
+  before_action :set_requsted_url
```
zachmargolis, line 5: typo: can we make this `set_requested_url` or maybe make it `return_to_url`?

the typo was copied very consistently

app/services/omniauth.rb
```diff
@@ -0,0 +1,25 @@
+class Omniauth
```
zachmargolis, line 1: Since omniauth is a gem....can we call this something else? Maybe `LoginHelpers` `SessionService` or something? This is not something I think we should monkeypatch into another namespace

slj: Didn't realize Omniauth was a gem, thanks for the alert.

PR #349, opened 2020-05-28
LG-3048: use case insensitive seach when adding emails to a team

solipet:
**Why?** Adding an existing user to the dashboard does not work with uppercase characters.

### DIFF
--- a/app/forms/manage_users_form.rb
+++ b/app/forms/manage_users_form.rb
@@ -23,9 +23,7 @@ class ManageUsersForm
   private
 
   def users_from_user_emails
-    existing_users = User.where(email: user_emails).to_a
-    missing_users = (user_emails - existing_users.map(&:email)).map { |e| User.new(email: e) }
-
+    missing_users = missing_emails.map { |e| User.new(email: e) }
     (existing_users + missing_users).sort_by(&:email)
   end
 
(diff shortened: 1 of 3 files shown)


solipet:
To be fair, the issue says it expects a warning that the user already exists, where this solution silently ignores the fact that you tried to add an existing user. If that's a problem, I can keep working on it.

jmhooper:
@solipet: I don't think that is a problem. We can stay with the existing behavior for the bug fix.

app/forms/manage_users_form.rb
```diff
@@ -35,4 +33,13 @@ def user_emails_are_valid_email_addresses
       errors.add(:base, "#{email} is not a valid email address")
     end
   end
+
+  def existing_users
+    User.where('lower(email) IN (?)', user_emails.map(&:downcase)).to_a
```
jmhooper, line 38: Messed up my review comment.

One think we could probably do here is downcase `user_emails` on submit:

```ruby
def submit(user_emails)
  @user_emails = user_emails.map(&:downcase)
end
```

I'll leave whether to do that up to you.

zachmargolis: What if we migrated the table to be `citext`? https://www.postgresql.org/docs/9.1/citext.html

solipet: Yes! That cleans it up considerably. I wrote this prior to discovering that devise was already downcasing emails on save.

solipet: @zachmargolis with devise already downcasing the emails, I don't want to mess with the existing schema/index. I tried to play with it, but couldn't quickly figure out how to get a rails migration to recognize the `citext` type.

zachmargolis: Ok as long as Devise is doing essentially the same thing, it's good then

PR #350, opened 2020-06-01
Add all agencies to config

achapm:


achapm:
Closing since this list includes IRS (Not a top level agency) and we want to pull from idp-config.

PR #352, opened 2020-06-02
Pull agency_id from team, so it correctly updates in JSON to the IDP

zachmargolis:
**Problem**

in INT dashboard `urn:gov:dhs.cbp.jobs:openidconnect:aws-goes-dev` and`urn:gov:dhs.cbp.pspd:openidconnect:pspd-ttp-dev` have the same agency/team, but in IDP they don't

The issue is that the IDP pulls `agency_id` out of JSON, which was being pulled as a direct attribute, but in dashboard this should come from the team association.

**Solution**

Fix the JSON serializer

### DIFF
--- a/spec/controllers/api/service_providers_controller_spec.rb
+++ b/spec/controllers/api/service_providers_controller_spec.rb
@@ -6,8 +6,10 @@ describe Api::ServiceProvidersController do
   end
 
   describe '#index' do
+    let(:team) { build(:team, agency: build(:agency)) }
+
     it 'returns active, approved SPs' do
-      sp = create(:service_provider, active: true, approved: true)
+      sp = create(:service_provider, active: true, approved: true, team: team)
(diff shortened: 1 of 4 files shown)


spec/controllers/api/service_providers_controller_spec.rb
```diff
@@ -6,8 +6,10 @@ def response_from_json
   end
 
   describe '#index' do
+    let(:team) { build(:team, agency: build(:agency)) }
+
     it 'returns active, approved SPs' do
-      sp = create(:service_provider, active: true, approved: true)
+      sp = create(:service_provider, active: true, approved: true, team: team)
```
solipet, line 12: There is a `:with_team` trait you could use from the sp factory

zachmargolis: That's great, will use. Thanks!

PR #355, opened 2020-06-05
LG-3046 Changed how Login.gov home link is announced

slj:
**Why**: It was being announced as HOME ALT+1 instead of Login.gov...
**How**: Removed accesskey attribute
**How**: Changed aria label to Login.gov partner dashboard
**How**: Changed title to Login.gov partner dashboard
**How**: Replaced user mailer slim templates with ERBs


slj:
https://cm-jira.usa.gov/browse/LG-3046

slj:
Thoughts, anyone?  Thx in advance.

app/views/layouts/application.html.erb
```diff
@@ -84,8 +84,7 @@
       <div class="usa-nav-container">
         <div class="usa-navbar">
           <div class="usa-logo" id="basic-logo">
-            <a href="/" accesskey="1" title="Home" aria-label="Home">
-              <%= image_tag 'img/login-gov-logo.svg', alt: 'login.gov', class: 'usa-logo__img' %>
+            <a href="/" title="Login.gov partner dashboard" aria-label="Login.gov partner dashboard">
               <em class="usa-logo__text"><%= Rails.application.config.app_name %></em>
```
zachmargolis, line 88: What is `Rails.application.config.app_name` -- is it different than the new `APP_NAME` constant?

config/application.rb
```diff
@@ -17,6 +17,8 @@
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
+APP_NAME = 'login.gov'.freeze
```
zachmargolis, line 20: Is this intended to be plain login.gov or the partner dashboard?

slj: It's intended to be 'login.gov'.  I created the constant while converting `app/views/user_mailer/welcome_new_user.html.slim` to `app/views/user_mailer/welcome_new_user.html.erb`.  I could change the constant name, but as it doesn't have much value beyond that template right now, maybe I should remove the constant and use a presenter?  It might be overkill, but it should avoid this confusion.  Thanks for pointing this out.

zachmargolis: I think that's misleading? This is the Dashboard app, why does the email say "welcome to login.gov?" Shouldn't it be "welcome to the login.gov integration dashboard" or something? login.gov sends its own emails, we need to make sure they're easy to differentiate, right? These are separate apps

slj: Where was my head?  The slim template was:
```
html
  head
    meta content='text/html; charset=UTF-8' http-equiv='Content-Type'
  body

  p Welcome to #{root_url}!

  p
    =link_to root_url, root_url

  p
    strong PLEASE DO NOT REPLY TO THIS MESSAGE
```
The link_to was what led me to talk to you about possibly doing something like `URI.parse(root_url).domain`. Now, I'm not sure about it.  I think I should just get rid of `APP_NAME` and keep `root_url` or change it to `root_path` (isn't that more appropriate for views), no?

zachmargolis: The other templates use `Rails.application.config.app_name` which is the Dashboard URL, so let's stick with that?

PR #357, opened 2020-06-09
Enable soft deletion of users

slj:
https://cm-jira.usa.gov/browse/LG-2897

Migration to prep users for soft-deletion

I'll make a separate PR for users' model/view/controller updates

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_05_22_215548) do
+ActiveRecord::Schema.define(version: 2020_06_09_184921) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -132,6 +132,7 @@ ActiveRecord::Schema.define(version: 2020_05_22_215548) do
     t.inet "last_sign_in_ip"
     t.boolean "admin", default: false, null: false
     t.integer "group_id"
+    t.datetime "deleted_at"
     t.index ["email"], name: "index_users_on_email", unique: true
     t.index ["group_id"], name: "index_users_on_group_id"
     t.index ["uuid"], name: "index_users_on_uuid", unique: true
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_06_09_184921) do
+ActiveRecord::Schema.define(version: 2020_06_11_125630) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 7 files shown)


db/schema.rb
```diff
@@ -132,6 +132,7 @@
     t.inet "last_sign_in_ip"
     t.boolean "admin", default: false, null: false
     t.integer "group_id"
+    t.datetime "deleted_at"
     t.index ["email"], name: "index_users_on_email", unique: true
```
jmhooper, line 136: This index may cause trouble. If we delete a user then try to add a user with the same email it will lead to an inconsistency.

We should be able to drop the index and then rebuild it with `WHERE deleted_at IS NOT NULL`.

jmhooper: I think the UUID index may need the same treatment. If a user get's deleted, changes their email in login, then gets re-added, that'll break things.

slj: Good catch - thank you.

slj: Wait...that should be `WHERE deleted_at IS NULL`, right?  We want the index to work on records that "exist", correct?

db/migrate/20200610152855_remove_email_uuid_unique_indeces_on_users.rb
```diff
@@ -0,0 +1,11 @@
+class RemoveEmailUuidUniqueIndecesOnUsers < ActiveRecord::Migration[5.2]
```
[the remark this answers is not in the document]
slj: Yes, I typed it wrong, it should have been "indices" (middle i), but I'm fine with "indexes"

db/migrate/20200610152855_remove_email_uuid_unique_indeces_on_users.rb
```diff
@@ -0,0 +1,11 @@
...
+  def change
```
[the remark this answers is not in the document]
slj: Earlier I was struggling with the rollback, maybe I didn't something wrong causing me to think I needed `down`, but I'll try this one more time and see if `change` suffices.

db/migrate/20200610152855_remove_email_uuid_unique_indeces_on_users.rb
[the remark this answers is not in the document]
slj: Yes, `change` is sufficient. Thanks again for calling this out.

slj:
If commit ed87686 is acceptable, I'll merge.  This sets up the next PR, where I plan to use the `acts_as_paranoid` gem (which uses the `deleted_at` column by default) for soft deletion.

zachmargolis:
I think this is good, but why separate the migrations from the code that adds the acts-as-paranoid gem? Migrations run before the new code starts up so it should be safe to do those together, right?

slj:
Since this PR didn't involve any code changes that would leverage the gem, I thought it was more appropriate for the next PR.  But I'm happy to add the gem here.

PR #358, opened 2020-06-12
Add soft deletion to User model

slj:
Surprisingly, the original User model validation:
```
validates :email, uniqueness: true
```
Does not leverage `act_as_paranoid` to add the `deleted_at` query scope:
Screen Shot 2020-06-11 at 9 13 28 AM

So I added a custom validator that works:
```
validates_with UserValidator, on: :create
```
Screen Shot 2020-06-11 at 9 17 59 AM

There are currently no Spanish/French translations in the app, so I didn't break the trend.  I just updated/normalized `config/locales/en.yml`.

### DIFF
--- a/app/validators/user_validator.rb
+++ b/app/validators/user_validator.rb
@@ -1,0 +1,7 @@
+class UserValidator < ActiveModel::Validator
+  def validate(record)
+    if User.find_by_email(record.email)
+      record.errors.add(:email, I18n.t('notices.user_already_exists'))
+    end
+  end
+end
--- a/app/validators/user_validator.rb
+++ b/app/validators/user_validator.rb
@@ -1,7 +1,6 @@
 class UserValidator < ActiveModel::Validator
   def validate(record)
-    if User.find_by_email(record.email)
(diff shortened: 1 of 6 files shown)


spec/models/user_spec.rb
```diff
@@ -71,4 +71,16 @@
       expect(user.scoped_teams).to eq(Team.all)
     end
   end
+
+  describe 'validates' do
+    it 'uniqueness of email address' do
```
zachmargolis, line 76: can we add a case where we delete a user and then are able to create a new account with that same email still?

app/validators/user_validator.rb
```diff
@@ -0,0 +1,7 @@
+class UserValidator < ActiveModel::Validator
+  def validate(record)
+    if User.find_by_email(record.email)
```
zachmargolis, line 3: I think we can use `exists?` for a slightly cheaper query (doesn't create a full AR object, doesn't load all columns frmo the DB)

```ruby
User.where(email: record.email).exists?
```

slj: Oh - this is good to know!  Thank you.

PR #359, opened 2020-06-15
Run job to clean users at configurable interval

slj:
Final PR for https://cm-jira.usa.gov/browse/LG-2897

### DIFF
--- a/app/jobs/clean_users_job.rb
+++ b/app/jobs/clean_users_job.rb
@@ -1,0 +1,9 @@
+class CleanUsersJob
+  include SuckerPunch::Job
+
+  def perform(interval)
+    CleanUsersService.call
+
+    CleanUsersJob.perform_in(interval, interval)
+  end
+end
--- a/app/services/clean_users_service.rb
+++ b/app/services/clean_users_service.rb
@@ -1,6 +1,6 @@
 class CleanUsersService
   def self.call
-    return unless User
+    return unless ActiveRecord::Base.connection
 
     count = User.where(last_sign_in_at: nil).
             where('created_at < ?', 14.days.ago).
--- a/app/services/clean_users_service.rb
+++ b/app/services/clean_users_service.rb
@@ -9,6 +9,6 @@ class CleanUsersService
     accounts = 'account'.pluralize(count)
     Rails.logger.info("Deleted #{count} unauthenticated #{accounts}")
   rescue ActiveRecord::NoDatabaseError
-    Rails.logger.info("Database not yet available")
+    Rails.logger.info('Database not yet available')
   end
 end
--- a/app/services/clean_users_service.rb
+++ b/app/services/clean_users_service.rb
@@ -9,5 +9,6 @@ class CleanUsersService
     accounts = 'account'.pluralize(count)
     Rails.logger.info("Deleted #{count} unauthenticated #{accounts}")
   rescue ActiveRecord::NoDatabaseError
+    Rails.logger.info("Database not yet available")
   end
 end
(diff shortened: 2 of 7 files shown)


slj:
My apologies for the excessive commits.  I MUST make a habit of running `make lint` until some of these gotchas are out of my system.

config/initializers/clean_users.rb
```diff
@@ -0,0 +1,7 @@
+lambda do
+  config = Rails.application.config
+  config.after_initialize do
+    interval = Figaro.env.clean_users_job_frequency_seconds || 30.days
+    CleanUsersJob.new.perform(interval.to_i)
+  end
+end.call
```
zachmargolis, line 7: what's the purpose fo doing this as an IIFE? Scoping? Local variables in Ruby are not accessible across files so I think it would be fine to do this without.

And the assigning to `config` is not needed either? This could all be:

```ruby
Rails.application.config.after_initialize do
  interval = Figaro.env.clean_users_job_frequency_seconds || 30.days
  CleanUsersJob.new.perform(interval.to_i)
end
```

app/services/clean_users_service.rb
```diff
@@ -0,0 +1,14 @@
+class CleanUsersService
+  def self.call
+    count = User.where(last_sign_in_at: nil).
+            where('created_at < ?', 14.days.ago).
+            delete_all
+
+    return unless count.positive?
+
+    accounts = 'account'.pluralize(count)
+    Rails.logger.info("Deleted #{count} unauthenticated #{accounts}")
```
zachmargolis, line 10: I think worrying about pluralizing is nice for UIs but not as useful for semi-structured logs?

What if the message was a little more machine-readable, then we could easily search/grep for `count=`

```
Rails.logger.info("Deleted unauthenticated accounts count=#{count}")
```

app/jobs/clean_users_job.rb
```diff
@@ -0,0 +1,9 @@
+class CleanUsersJob
+  include SuckerPunch::Job
+
+  def perform(interval)
+    CleanUsersService.call
+
+    CleanUsersJob.perform_in(interval, interval)
```
zachmargolis, line 7: Are there any other alternatives to scheduling recurring jobs like this? My worry is that if one instance of the job blows up, the jobs are going to stop until the next time the process gets rebooted.

A few ideas:
1. Could we use a crontab to schedule this? Something like the `whenever` gem?
2. If not, could we do this next `perform_in` in an `ensure` block so that even if the job blows up, we still schedule the next one?

jmhooper: IIRC in planning we talked about making this a button or something similar so we didn't have to worry about tooling for a job.

config/initializers/clean_users.rb
slj: I had an earlier revision where I thought it made sense, but clearly I didn't clean it up well enough.  Thanks for the pointers.

app/services/clean_users_service.rb
slj: Got it.

app/jobs/clean_users_job.rb
slj: I'm happy to do what you feel is best.  I was admittedly looking for a solution that avoided an external dependency like `crontab`, but I didn't take into account the possibility of the service failure preventing future events.  Thank you very much for pointing this out.  If you're agreeable, I would like to see how well `ensure` addresses your concern, but again, I'm willing to forgo the whole approach if you feel strongly otherwise.

slj: My apologies @hooper I didn't see your comment before I wrote mine. It sounds like the button is the agreed approach - I'll update the AC.  Question - shall all dashboard admins be allowed to click this button?  Shall we instead restrict it to a few specific users?

slj:
Closing this PR after confirming button approach with @porta-antiporta

PR #360, opened 2020-06-16
Admins can delete unconfirmed users (LG-2897)

slj:


app/controllers/unconfirmed_users_controller.rb
```diff
@@ -0,0 +1,8 @@
+class UnconfirmedUsersController < ApplicationController
+  def delete
```
zachmargolis, line 2: 1. We need to add the `before { authorize user }` filter here to make sure the policy gets used, right? 
2. The built-in rails action for deleting is `destroy` not `delete` IIRC --- but the HTTP verb is `delete` (not POST) so it gets kind of confusing

app/controllers/unconfirmed_users_controller.rb
```diff
@@ -0,0 +1,8 @@
...
+    count = CleanUsersService.call
```
stevegsa, line 3: Love the new service! <3  How about a more descriptive name like `DeleteUnconfirmedUsers`?

app/controllers/unconfirmed_users_controller.rb
slj: I'll look into the before_action, as for the HTTP verb, if I change the method to destroy, would that make more sense with POST?  Thanks for the quick response.

app/controllers/unconfirmed_users_controller.rb
slj: Thx for the quick response - so you mean change the controller name?  If so, is `DeleteUnconfirmedUsersController` ok?

app/controllers/unconfirmed_users_controller.rb
zachmargolis: I think that either
- DELETE verb + #destroy makes sense
- POST verb + #create makes sense... because we're "creating" an instance of the delete, if that makes sense?

zachmargolis: Reading more closely, HTTP DELETE is supposed to idempotent (because it usually refers to a specific resource, not a collection) so I think if our goal is something that is *not* retryable, then POST makes more sense? 🤷 

I don't feel that strongly which verb it is, I just think we should aim to have the verb + action name match if possible

slj: Adding `before_action -> { authorize current_user }` to the controller somehow casues  this error we saw before with UserPolicy:
```
NoMethodError at /remove_unconfirmed_users
undefined method `delete?' for #<UserPolicy:0x00007fcc1a482e50>
```
not sure what's going on...investigating.

zachmargolis: The NoMethodError is the same as yesterday -- the `UserPolicy` has methods for the built-in Rails actions (`create`, `new`, `destroy`, etc etc) so that's one more reason to use one of those methods as the action name

slj: Ok, I think I understand what's going on.  The UserPolicy needed a `delete?` method to tell the app, "this user is authorized to run the delete method on the controller".  I should be able to change the controller method to `destroy` - the User policy already has a `destroy?` method that is true for admins, so this should work.

app/controllers/unconfirmed_users_controller.rb
slj: Ohhhhh.....I think you mean the service name!  On it...

app/views/users/index.html.erb
```diff
@@ -1,25 +1,47 @@
...
+      <tr>
+        <th scope="col">Email</th>
+        <th scope="col">Signed in</th>
+        <th scope="col">Admin</th>
+        <th scope="col">Actions</th>
+      </tr>
+    </thead>
+    <tbody>
+      <% @users.each do |user| %>
+      <tr>
```
zachmargolis, line 27: small nit: can we indent this `<tr>` past the `.each` loop? Helps with matching

PR #364, opened 2020-06-19
Cleanup partner self-service onboarding

jmhooper:
**Why**: When I put together the manage users story I left some things off

**How**:
- Fix a 500 that prevents people from creating a team after the manage users change
- Redirect to the "Manage Users" page after creating a team

### DIFF
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -70,7 +70,7 @@ class TeamsController < AuthenticatedController
       team_params
     else
       user_ids = team_params[:user_ids] || []
-      team_params.merge('user_ids': (user_ids << current_user.id.to_s))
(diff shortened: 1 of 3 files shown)


app/controllers/teams_controller.rb
```diff
@@ -70,7 +69,8 @@ def update_params_with_current_user
     if current_user.admin?
       team_params
     else
-      team_params.merge('user_ids': (team_params[:user_ids] << current_user.id.to_s))
+      user_ids = team_params[:user_ids] || []
+      team_params.merge('user_ids': (user_ids << current_user.id.to_s))
```
zachmargolis, line 73: nit: seems weird to me to quote the key (I'd expect `user_ids:` instead of `'user_ids':`

jmhooper: Agree, willfix

PR #368, opened 2020-07-28
RP-initiated logout (LG-3050)

mitchellhenke:
Also fixes a couple small issues I had when setting up the application. Agencies being seeded were being inserted with specific IDs, which caused an error when an agency was being added later during setup:

```
PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "agencies_pkey"
DETAIL:  Key (id)=(1) already exists.
```

And removed package-lock.json since yarn complains that the file exists.

Pending the changes in https://github.com/18F/omniauth_login_dot_gov/pull/13


spec/controllers/users/omniauth_controller_spec.rb
```diff
@@ -10,6 +10,9 @@
           'email' => email,
           'uuid' => uuid,
         },
+        'credentials'=> {
```
zachmargolis, line 13: super nit, can we fix the spacing here? To add a space before the hashrocket here and on the next line? ` =>`

app/controllers/users/omniauth_controller.rb
```diff
@@ -6,10 +6,15 @@ def callback
...
       if @user
         sign_in @user
+        set_id_token
         redirect_to session[:requested_url]
       else
         redirect_to users_none_url
       end
     end
+
+    def set_id_token
```
zachmargolis, line 16: maybe call it `store_id_token` ? `set` makes me think of an instance variable setter

app/controllers/users/sessions_controller.rb
```diff
@@ -22,9 +22,21 @@ def timeout
...
-      # TODO: log out of IdP using OIDC logout endpoint
-      #   https://developers.login.gov/oidc/#logout
-      super
+      if session[:id_token]
+        logout_request = self.class.logout_utility.build_request(
+          id_token: session[:id_token],
+          post_logout_redirect_uri: 'http://localhost:3001/'
+        )
+        sign_out(current_user)
+        redirect_to(logout_request[:redirect_uri]) and return
```
zachmargolis, line 31: since this is the last line in the branch, and there's no other code, the `and return` is redundant right?

zachmargolis: Also can we use the property name? `logout_request.redirect_uri`

app/controllers/users/sessions_controller.rb
```diff
@@ -22,9 +22,21 @@ def timeout
     end
 
     def destroy
-      # TODO: log out of IdP using OIDC logout endpoint
-      #   https://developers.login.gov/oidc/#logout
-      super
+      if session[:id_token]
+        logout_request = self.class.logout_utility.build_request(
+          id_token: session[:id_token],
+          post_logout_redirect_uri: 'http://localhost:3001/'
```
zachmargolis, line 28: we should pull this from a config or env var

app/controllers/users/sessions_controller.rb
mitchellhenke: Thanks for catching that. I was having issues with ordering of signing out/redirect and the `and return` was necessary if you wanted to do something after a redirect, but the if/else moved things around such that it wasn't needed.

zachmargolis:
Looks like there are some Snyk security issues.... if we've already fixed these on the `master` branch syncing with master should fix. Otherwise maybe upgrading rails-webpacker npm package?

mitchellhenke:
> Looks like there are some Snyk security issues.... if we've already fixed these on the `master` branch syncing with master should fix. Otherwise maybe upgrading rails-webpacker npm package?

The branch should be up to date with master. I don't have access to see the specific snyk issues, but it may be related to the omniauth CVE?

PR #377, opened 2020-10-01
Fix flaky test

mitchellhenke:
This test sometimes fails because the arrays have the same elements in a different order

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -33,7 +33,8 @@ describe User do
       user_sp = create(:service_provider, user: user)
       team_sp = create(:service_provider, team: team)
       create(:service_provider)
-      expect(user.scoped_service_providers).to eq([user_sp, team_sp])
+      sorted_sps = [user_sp, team_sp].sort_by { |x| x.friendly_name.downcase }
+      expect(user.scoped_service_providers).to eq(sorted_sps)
     end
     it "alphabetizes the list of user created and the user's team sps" do
       team = create(:team)
--- a/yarn.lock
+++ b/yarn.lock
@@ -900,17 +900,11 @@
   resolved "https://registry.yarnpkg.com/@sinonjs/text-encoding/-/text-encoding-0.7.1.tgz#8da5c6530915653f3a1f38fd5f101d8c3f8079c5"
   integrity sha512-+iTbntw2IZPb/anVDbypzfQa+ay64MW0Zo8aJ8gZPWMMK6/OubMVb6lUPMagqjOPnmtauXnFCACVl3O7ogjeqQ==
 
-"@types/events@*":
-  version "3.0.0"
-  resolved "https://registry.yarnpkg.com/@types/events/-/events-3.0.0.tgz#2862f3f58a9a7f7c3e78d79f130dd4d71c25c2a7"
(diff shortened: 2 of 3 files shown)


aduth (approved):
:+1:

I thought this might be a problem since the method [does expect a sort order](https://github.com/18F/identity-dashboard/blob/5460756ea8520ebbd63dac865f475b1298d531e9/app/models/user.rb#L21-L25), but I see there's a test case below which guarantees this.

I imagine the flakiness could be something in [how we generate service providers](https://github.com/18F/identity-dashboard/blob/5460756ea8520ebbd63dac865f475b1298d531e9/spec/factories/service_providers.rb#L3), e.g. `["test98","test99","test100"].sort == ["test100", "test98", "test99"]`.

spec/models/user_spec.rb
```diff
@@ -33,7 +33,8 @@
       user_sp = create(:service_provider, user: user)
       team_sp = create(:service_provider, team: team)
       create(:service_provider)
-      expect(user.scoped_service_providers).to eq([user_sp, team_sp])
+      sorted_sps = [user_sp, team_sp].sort_by { |x| x.friendly_name.downcase }
+      expect(user.scoped_service_providers).to eq(sorted_sps)
```
zachmargolis, line 37: We can use RSpec's `match_array` to simplify this (it does order-independent comparison)

```suggestion
      expect(user.scoped_service_providers).to match_array([user_sp, team_sp])
```

some examples: https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers

yarn.lock
```diff
@@ -900,17 +900,11 @@
...
-  resolved "https://registry.yarnpkg.com/@types/events/-/events-3.0.0.tgz#2862f3f58a9a7f7c3e78d79f130dd4d71c25c2a7"
-  integrity sha512-EaObqwIvayI5a8dCzhFrjKzVwKLxjoG9T6Ppd5CEo07LRKfQ8Yokw54r5+Wq7FaBQ+yXRvQAYPrHwya1/UFt9g==
-
 "@types/glob@^7.1.1":
-  version "7.1.1"
-  resolved "https://registry.yarnpkg.com/@types/glob/-/glob-7.1.1.tgz#aa59a1c6e3fbc421e07ccd31a944c30eba521575"
-  integrity sha512-1Bh06cbWJUHMC97acuD6UMG29nMt0Aqz1vF3guLfG+kHHJhy3AyohZFFxYk2f7Q1SQIrNwvncxAE0N/9s70F2w==
+  version "7.1.3"
+  resolved "https://registry.yarnpkg.com/@types/glob/-/glob-7.1.3.tgz#e6ba80f36b7daad2c685acd9266382e68985c183"
+  integrity sha512-SEYeGAIQIQX8NN6LDKprLjbrd5dARM5EXsd8GI/A5l0apYI1fGMWgPHSe4ZKL4eozlAyI+doUE9XbYS4xCkQ1w==
```
zachmargolis, line 906: Do we need all these yarn updates too?

zachmargolis (approved):
LGTM, one suggestion

yarn.lock
mitchellhenke: I was trying to sneak in the vulnerability fixes 😬

spec/models/user_spec.rb
mitchellhenke: That's what I had initially, but Andrew caught that `scoped_service_providers` is actually sorted by `friendly_name`, so I changed the test to match that behavior

zachmargolis: Oh ok! sounds good

PR #378, opened 2020-11-27
LG-3771 Upgrade to Rails 6

stevegsa:


config/environments/development.rb
```diff
@@ -57,7 +57,7 @@
   # Debug mode disables concatenation and preprocessing of assets.
   # This option may cause significant delays in view rendering with a large
   # number of complex assets.
-  config.assets.debug = true
+  config.assets.debug = false
```
mitchellhenke, line 60: I think this should still default to true? https://github.com/rails/rails/blob/3446efc8bca2cb310aabc4b2e801598252344e0e/guides/source/configuring.md#configuring-assets

mitchellhenke (commented):
Couple questions 🙂

app/models/service_provider.rb
```diff
@@ -107,7 +107,7 @@ def logo_file_mime_type
     return unless logo_file.attached?
     return if mime_type_valid?
     errors.add(:logo_file, "The file you uploaded (#{logo_file.filename}) is not a PNG or SVG")
-    logo_file.purge
+    logo_file = nil # rubocop:disable Lint/UselessAssignment
```
mitchellhenke, line 110: can we remove the line instead of doing the assignment? I'm not super familiar with how the logo files get stored, does it have something to do with that?

mitchellhenke (approved):
One question, otherwise 👍🏼

app/models/service_provider.rb
stevegsa: Yes: https://github.com/rails/rails/issues/37069#issuecomment-532462387

mitchellhenke: 👍🏼

PR #379, opened 2020-11-30
[LG-2898] Allow admins to enable prompt=login for SPs

orenyk:
Add boolean allow_prompt_login attribute to the service_providers table
that defaults to false, matching what the IdP has. Conditionally add
checkbox field for admins on the service provider form to allow editing
of that attribute, and also authorize editing of that attribute at the
controller level. Finally, add display for this attribute in the show
view.

Includes a feature spec for the happy path and unit tests for a new
helper method for the show view.

### DIFF
--- a/app/views/service_providers/_form.html.erb
+++ b/app/views/service_providers/_form.html.erb
@@ -73,6 +73,20 @@
     <%= form.input :production_issuer,
                    label: 'Production Issuer',
                    input_html: { class: 'usa-input' } %>
+
+    <%# allow_prompt_login checkbox %>
+    <fieldset class="usa-fieldset">
+      <legend class="usa-sr-only">Enable prompt=login</legend>
+      <%= form.label :allow_prompt_login,
+                      label: 'Enable <code>prompt=login</code>'.html_safe,
+                      class: 'usa-label' %>
+      <ul class="usa-input-list">
+        <li>
+          <%= form.check_box :allow_prompt_login, class: 'usa-checkbox__input' %>
+          <%= form.label :allow_prompt_login, 'Enabled', class: 'usa-checkbox__label' %>
+        </li>
+      </ul>
+    </fieldset>
   <% end %>
 
   <%# Temp hack to allow feature in dev but not int for testing %>
--- a/spec/features/service_providers_spec.rb
+++ b/spec/features/service_providers_spec.rb
@@ -101,6 +101,16 @@ feature 'Service Providers CRUD' do
       expect(service_provider.redirect_uris).to eq(['https://bar.com'])
     end
 
+    scenario 'cannot edit allow_prompt_login' do
+      user = create(:user, :with_teams)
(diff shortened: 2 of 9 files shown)


orenyk:
Note that the build is failing because the node version in Circle CI is v14 and it requires a max of v12.

aduth:
> Note that the build is failing because the node version in Circle CI is v14 and it requires a max of v12.

For context, this is how it was fixed in `identity-idp`: https://github.com/18F/identity-idp/pull/4353

We might want to update similarly in this repository, though not necessarily as part of this pull request.

app/views/service_providers/_form.html.erb
```diff
@@ -73,6 +73,20 @@
     <%= form.input :production_issuer,
                    label: 'Production Issuer',
                    input_html: { class: 'usa-input' } %>
+
+    <%# allow_prompt_login checkbox %>
```
achapm, line 77: Should we display a message "Prompt=login enabled" message for non-admin users after prompt=login has been enabled?

spec/features/service_providers_spec.rb
```diff
@@ -197,6 +197,18 @@
...
       expect(page).to have_content(I18n.t('notices.service_providers_refreshed'))
     end
+
+    scenario 'can enable prompt=login for a service provider' do
+      admin = create(:admin)
+      sp = create(:service_provider, :with_team)
+      login_as(admin)
+
+      visit edit_service_provider_path(sp)
+      check 'Enabled'
```
achapm, line 207: Suggestion...
Should we address potential conflicts with other checkbox's added to this page?

Should we try 
```
check 'allow_prompt_login'
```
And reference the ID of the field? If `Enabled` is the field ID, then maybe we should change it to avoid elements with the same id.

spec/features/service_providers_spec.rb
```diff
@@ -197,6 +197,18 @@
       click_on t('forms.buttons.trigger_idp_refresh')
       expect(page).to have_content(I18n.t('notices.service_providers_refreshed'))
     end
+
```
achapm, line 200: I suggest we add a feature test to insure the checkbox does not display for non admins

app/views/service_providers/_form.html.erb
orenyk: Good question. I followed the example set with other admin-only attributes and made the display on the show view admin-only as well. I'll add a separate branch for users and we'll only show it if it's turned on.

spec/features/service_providers_spec.rb
orenyk: Done!

spec/features/service_providers_spec.rb
orenyk: Yup makes sense, that was just the label so I updated the test.

achapm (approved):
LGTM

PR #381, opened 2020-12-11
Fix rack timeout

stevegsa:


config/application.yml.example
```diff
@@ -1,5 +1,6 @@
 # These settings are common across environments, but may be overridden
 # in any particular environment
+RACK_TIMEOUT_SERVICE_TIMEOUT: '15'
```
mitchellhenke, line 3: Should we add the same note we have in the IDP since this config is a bit of an odd case?
https://github.com/18F/identity-idp/blob/master/config/application.yml.default#L18-L21

mitchellhenke (approved):
Looks good to me 👍🏼

PR #383, opened 2020-12-22
[LG-3686] Validate that at least one user is added to a team

orenyk:
Resolves LG-3686

Adds a validation to the ManageUsersForm to check that at least one
email address is submitted. Also updates ManageUsersController#create to
pass in an empty array instead of nil when no emails are passed.


achapm (approved):
LGTM

PR #385, opened 2021-01-06
[LG-3733] Add error page for 413 Request Entity Too Large

orenyk:
Adding a custom error page that indicates that this error may be representative of an issue with their logo file size.


public/413_logo.html
```diff
@@ -0,0 +1,121 @@
...
+                </svg>
+              </div>
+            </div>
+          </div>
+          <div class="inner cover">
+            <h1>We're sorry, but your request was too large.</h1>
+            <p>
+            If you were trying to upload a logo for your app, please review our
+            <a href="https://developers.login.gov/design-guidelines/#agency-logo-guidelines" target="_blank">guidelines</a>
+            and make sure the image is no larger than 1MB. You can also go back
```
zachmargolis, line 112: Do we have any JS that can detect the file size before we submit the form to the server?

zachmargolis (approved):
LGTM

public/413_logo.html
orenyk: I think that's the next step - right now we're just trying to mitigate the extremely unhelpful 413 error page that nginx throws when you upload a logo that's too big. The other option is to enable multi-part uploads to mitigate the performance and DDOS risk of large request sizes, but we wanted to try and get something quick and low-risk done first.

jgrevich:
Nice! I'll make a follow up PR to add the needed changes for this on the devops side: https://gsa-tts.slack.com/archives/C16RSBG49/p1609965121018600?thread_ts=1609960015.011800&cid=C16RSBG49

jgrevich (approved):
👍 nitpick, ideally the style should match the IdP's error pages

jgrevich:
The style looks good to me now
Screen Shot 2021-01-06 at 7 41 11 PM

Perhaps it as an issue with me loading the html file from my desktop.

PR #386, opened 2021-01-08
[LG-3733] Add client-size validation of logo size

orenyk:
Warn partners if they select a file greater than 1MB in size. We could
also add all sorts of other client-side validations but keeping it
simple for now.

### DIFF
--- a/app/javascript/packs/validate-logo-size.js
+++ b/app/javascript/packs/validate-logo-size.js
@@ -1,0 +1,23 @@
+document.getElementById("service_provider_logo_file").addEventListener("change", (e) => {
+  // Clear the error div
+  var errorDiv = document.getElementById("logo-upload-error");
+  errorDiv.textContent = "";
+
+  // See https://stackoverflow.com/a/3717847
+  if (!window.FileReader) {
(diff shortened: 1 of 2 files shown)


app/javascript/packs/validate-logo-size.js
```diff
@@ -0,0 +1,23 @@
+document.getElementById("service_provider_logo_file").addEventListener("change", (e) => {
+  // Clear the error div
+  error_div = document.getElementById("logo-upload-error");
```
zachmargolis, line 3: Let's
1. declare the variable to scope it
2. use camelCase because it's JS
```suggestion
  var errorDiv = document.getElementById("logo-upload-error");
```

zachmargolis (approved):
LGTM

app/javascript/packs/validate-logo-size.js
orenyk: was it that obvious that my JS brain is still evolving? 😛

orenyk: Fixed!

PR #388, opened 2021-01-25
Adds default_aal to service_provider (LG-3825)

solipet:
Allow partners to set the default AAL level via the dashboard. This value is used if no AAL value is specified explicitly in the request.

Adds an select input to the SP form to choose a level, and shows it when viewing the SP.

### DIFF
--- a/app/models/service_provider.rb
+++ b/app/models/service_provider.rb
@@ -41,10 +41,8 @@ class ServiceProvider < ApplicationRecord
     case default_aal
     when 1, nil
       ''
-    when 2
-      'AAL2'
-    when 3
-      'AAL3'
+    when 2, 3
+      "AAL#{default_aal}"
     else
       default_aal.inspect
     end
--- a/app/models/service_provider.rb
+++ b/app/models/service_provider.rb
@@ -37,6 +37,19 @@ class ServiceProvider < ApplicationRecord
     end
   end
 
+  def aal_friendly
+    case default_aal
+    when 1, nil
+      ''
+    when 2
+      'AAL2'
+    when 3
+      'AAL3'
+    else
+      default_aal.inspect
+    end
+  end
+
   # rubocop:disable MethodLength
   def self.possible_attributes
     possible = %w[
--- a/app/serializers/service_provider_serializer.rb
+++ b/app/serializers/service_provider_serializer.rb
@@ -9,6 +9,7 @@ class ServiceProviderSerializer < ActiveModel::Serializer
     :cert,
     :friendly_name,
     :ial,
+    :default_aal,
     :issuer,
     :logo,
     :remote_logo_key,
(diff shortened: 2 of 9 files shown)


db/migrate/20210122185518_add_default_aal_to_service_providers.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultAalToServiceProviders < ActiveRecord::Migration[6.0]
+  def change
+    add_column :service_providers, :default_aal, :integer
```
zachmargolis, line 3: do we need to add the same aal -> default_aal copy as we have in the IDP PR?

app/serializers/service_provider_serializer.rb
```diff
@@ -9,6 +9,7 @@ class ServiceProviderSerializer < ActiveModel::Serializer
     :cert,
     :friendly_name,
     :ial,
+    :default_aal,
```
zachmargolis, line 12: can we add specs for this class too? or the controller that serves this serializer?

zachmargolis (approved):
LGTM

app/models/service_provider.rb
```diff
@@ -37,6 +37,17 @@ def ial_friendly
...
   end
 
+  def aal_friendly
+    case default_aal
+    when 1, nil
+      ''
+    when 2, 3
+      "AAL#{default_aal}"
+    else
+      default_aal.inspect
```
mitchellhenke, line 47: I was wondering what case would lead to a `default_aal` that wasn't `nil, 1, 2, 3`, should we have a validation?

```ruby
validates :default_aal, numericality: { only_integer: true, allow_nil: true, less_than_or_equal_to: 3 }
```

mitchellhenke (approved):
One question, otherwise 👍🏼

db/migrate/20210122185518_add_default_aal_to_service_providers.rb
solipet: Nope - there isn't an existing `aal` field on the dashboard 🙂

app/serializers/service_provider_serializer.rb
solipet: Are you looking for more than this?
https://github.com/18F/identity-dashboard/pull/388/files#diff-b698ab666b7c150a9fb27a9931adbf4ca9e203ae16ac80a728fcf1d24dbff068R29-R30

db/migrate/20210122185518_add_default_aal_to_service_providers.rb
orenyk: Are we always expecting this to be needed, e.g. should we have a `null: false` and a default set?

orenyk (approved):
Small question, otherwise looks good!

db/migrate/20210122185518_add_default_aal_to_service_providers.rb
solipet: `null` is fine - currently, only two SPs have anything set at all in the existing `aal` field.

app/models/service_provider.rb
solipet: I don't feel strongly, but if we wanted to add the validation, it would go in the `identity-validations` gem.

app/serializers/service_provider_serializer.rb
zachmargolis: Nope! I must have missed that, thanks!

PR #389, opened 2021-01-26
Improve first-run documentation guidance

aduth:
**Why**: As a developer contributing to identity-dashboard, I want to know what to do when I log in for the first time and encounter a "No account exists" warning page, so that I'm not left confused and forced to resort to scouring through source code to figure out how to use the application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,10 @@ These instructions assume [`identity-idp`](https://github.com/18F/identity-idp) 
 
 In the development environment, an administrator user account is created during the setup process with the email address _admin@gsa.gov_ . If this is the first time you've run the application, you will want to log in with this email address, or create an account if one doesn't already exist. After you've logged in as the administrator, you can add new users and teams.
 
+Alternatively, log in using any account, then promote the user to an admin using the `users:make_admin` Rake task:
+
+`bundle exec rake users:make_admin USER=user@example.com,First,Last`
+
 ## Running tests
 
 Run RSpec tests using:
--- a/README.md
+++ b/README.md
@@ -26,6 +26,8 @@ These instructions assume [`identity-idp`](https://github.com/18F/identity-idp) 
   $ make run
   ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,6 +26,8 @@ These instructions assume [`identity-idp`](https://github.com/18F/identity-idp)
   $ make run
   ```
 
+In the development environment, an administrator user account is created during the setup process with the email address _admin@gsa.gov_ . If this is the first time you've run the application, you will want to log in with this email address, or create an account if one doesn't already exist. After you've logged in as the administrator, you can add new users and teams.
```
zachmargolis, line 29: another option is to log in with anything and just go into the rails console and set any other user's `user.admin = true; user.save`

zachmargolis (approved):
LGTM

README.md
aduth: > another option is to log in with anything and just go into the rails console and set any other user's `user.admin = true; user.save`

Yeah, I found there's a couple additional options, like...

- Signing in with a government email address [creates a new (non-admin) user](https://github.com/18F/identity-dashboard/blob/7a1f7a7be011f54abf505330fb3e66f315466f6c/app/services/user_session.rb#L16-L24)
- Using [one of the existing Rake tasks](https://github.com/18F/identity-dashboard/blob/master/lib/tasks/users.rake) to promote a user to admin
- Using Rails console to create a new user manually (`User.new(email: '...', admin: true)`)

Wasn't sure how much of this we think is important to document, vs. bare minimum of getting someone up-and-running through the browser alone.

aduth: Maybe something like...

```suggestion
In the development environment, an administrator user account is created during the setup process with the email address _admin@gsa.gov_ . If this is the first time you've run the application, you will want to log in with this email address, or create an account if one doesn't already exist. After you've logged in as the administrator, you can add new users and teams.

Alternatively, log in using any account, then promote the user to an admin using the `users:make_admin` Rake task:

`bundle exec rake users:make_admin USER=user@example.com,First,Last`
```

zachmargolis: Oh right, the rake task! Seems worthwhile including that

aduth: 👍  Updated in 23ffcb7.

PR #390, opened 2021-01-29
HOTFIX: save blank AAL as nil, not 1

solipet:
This results in us sending back `http://idmanagement.gov/ns/assurance/aal/1` instead of `urn:gov:gsa:ac:classes:sp:PasswordProtectedTransport:duo`

### DIFF
--- a/app/views/service_providers/_form.html.erb
+++ b/app/views/service_providers/_form.html.erb
@@ -56,7 +56,7 @@
   <%= form.input :default_aal,
                  as: :select,
                  collection: [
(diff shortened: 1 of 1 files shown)


zachmargolis (approved):
LGTM, would love a spec 🙏

solipet:
> LGTM, would love a spec 🙏

Yup - I wrote a spec, seemed simple enough, but it didn't actually test the field. This is affecting partners now, so I merged the fix (testing manually locally) and will sort out a spec soon.

PR #392, opened 2021-02-22
Use lograge for structured logs (LG-4156)

zachmargolis:
Matches what we have for the IDP

Doesn't add an `events.log` like we have in the IDP, I am not sure that's needed yet


mitchellhenke (approved):

stevegsa (approved):
Looks good!

PR #393, opened 2021-02-26
Structured logging redo, use identity-logging gem

zachmargolis:
This is a redo of: https://github.com/18F/identity-dashboard/pull/392 to fix log formatting (forgot to copy over UpayaLogFormatter from the IDP) to have some more hands-free configuration via a new version of identity-hostdata, see https://github.com/18F/identity-hostdata/pull/15

~However, it gets a little spicier because I found that in order to include some extra info by default in that gem's config, I needed access to `payload[:request]` which is only found in Rails 6.1.... so I opted to include the Rails 6.1 upgrade~

~If anybody feels strongly, happy to separate out that change.~

~There are also some questions in that PR about implicit vs explicit railtie stuff, so I'm happy to wait for that to get sorted out before merging this~

**Updates**:
1. I separated out Rails 6.1 as its own PR: https://github.com/18F/identity-dashboard/pull/394, view the [incremental diff](https://github.com/18F/identity-dashboard/compare/margolis-rails-6-1...margolis-hostdata-logging) to see just these changes
2. The logging config now comes from its own gem, not from identity-hostdata


Gemfile
```diff
@@ -14,7 +14,7 @@ gem 'dotenv-rails', '~> 2.7', '>= 2.7.6'
 gem 'enum_help'
 gem 'figaro'
 gem 'httparty'
-gem 'identity-hostdata', git: 'https://github.com/18F/identity-hostdata.git', branch: 'master'
+gem 'identity-hostdata', git: 'https://github.com/18f/identity-hostdata.git', branch: 'margolis-railtie-with-logger'
```
zachmargolis, line 17: TODO:
- [ ] use a tagged version instead of pointing it a branch

stevegsa (approved):
LGTM

db/schema.rb
```diff
@@ -67,17 +74,7 @@
     t.index ["name"], name: "index_groups_on_name", unique: true
   end
 
-  create_table "help_texts", force: :cascade do |t|
```
mitchellhenke, line 70: should this have been dropped?

zachmargolis: I think that it was already dropped but left around lingering in this file

zachmargolis:
```
git grep help_text -- db/migrate
```

only shows column names for this, not a separate table

zachmargolis:
Update: I split out Rails 6.1 here: https://github.com/18F/identity-dashboard/pull/394

and updated this PR

PR #399, opened 2021-03-23
Change the newrelic.yml to match other repos and prod config

jmhooper:
**Why**: So dashboard can start managing its own NewRelic.yml instead of depending on the chef to write it one

### DIFF
--- a/config/newrelic.yml
+++ b/config/newrelic.yml
@@ -1,34 +1,21 @@
-common: &default_settings
-  app_name: "identity-dashboard"
+production:
+  agent_enabled: true
+  app_name: pivcac.<%= Identity::Hostdata.env %>.<%= Identity::Hostdata.domain %>
+  host: gov-collector.newrelic.com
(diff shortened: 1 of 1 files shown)


mitchellhenke (approved):

config/newrelic.yml
```diff
@@ -1,34 +1,21 @@
-common: &default_settings
-  app_name: "identity-dashboard"
+production:
+  agent_enabled: true
+  app_name: pivcac.<%= Identity::Hostdata.env %>.<%= Identity::Hostdata.domain %>
```
zachmargolis, line 3: this is the dashboard repo... but this says `pivcac`?

jmhooper: Got busted copy/pasting. How embarrassing. Willfix

jmhooper: Fix: https://github.com/18F/identity-dashboard/pull/400/files

PR #366, opened 2020-07-06
Fixed inconsistent headings

slj:
https://cm-jira.usa.gov/browse/LG-3138

Added `axe-matchers` gem and `features/accessibility` specs.  The gem found other issues besides missing headings, including:

- Portions of markup that were not in a specific page region, like `header` or `nav`
- Out-of-sequence headings (like `h1` followed by `h4` instead of `h2`

PR #404, opened 2021-03-30
Certificate cleanup

zachmargolis:
I was starting to make changes to the dashboard to support multiple to correspond to https://github.com/18F/identity-idp/pull/4851

But then I noticed a lot of things I wanted to clean up so I started on that:

1. Removed `_certificate_expiration.html.erb` partial
    - Inline styles are blocked by our CSP
    - Moved CSS logic on to `ServiceProviderCertificate` class
2. Updated `ServiceProviderCertificate` class, it now wraps an X509 cert rather than subclasses one (easier to test)
3. Visual changes: updated the SP "show" page:
    - Full cert body is now hidden behind a details toggle
        -  ~I updated USWDS to fix an error with Chrome and the triangle disclosure being hidden~ (reverted that fix based on https://github.com/18F/identity-dashboard/pull/404#discussion_r604372425)
    - Added some more summary details (I expect these will make it easier to tell multiple certificates apart)

| state | before | after |
| ---- | ---- | ---- |
| initial | Screen Shot 2021-03-30 at 10 17 42 AM | Screen Shot 2021-03-30 at 10 17 57 AM
| expanded | N/A | Screen Shot 2021-03-30 at 10 18 03 AM |


app/models/service_provider_certificate.rb
```diff
@@ -18,4 +34,12 @@ def self.expiration_time_to_colorized_s(time)
   def self.warning_period
     (Figaro.env.certificate_expiration_warning_period || 60).to_i.days.from_now
   end
+
+  def expiration_css_class
+    if not_after < Time.zone.now
+      'certificate-expired'
+    elsif not_after < self.class.warning_period
+      'certificate-warning'
+    end
```
aduth, line 43: If we want to retain the previous colors while avoiding the inline style, maybe we could use [design system color utility classes](https://designsystem.digital.gov/utilities/color/) and avoid the custom classes? This would also give us some color consistency for free. 

```suggestion
    if not_after < Time.zone.now
      'bg-red text-white'
    elsif not_after < self.class.warning_period
      'bg-yellow text-ink'
    end
```

package.json
```diff
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "@rails/webpacker": "^4.3.0",
-    "identity-style-guide": "^2.2.3",
+    "identity-style-guide": "^5.0.3",
```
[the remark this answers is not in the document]
zachmargolis: Went with the patch approach in 5c6741f791de77308be4d8fc0c740a5651f4321d

app/views/service_providers/show.html.erb
```diff
@@ -58,27 +58,48 @@
...
 
-<h2><label for="block_encryption">SAML Assertion Encryption:</label></h2>
-<p class="font-mono-xs margin-top-0" name="block_encryption"><%= service_provider.block_encryption %></p>
-    <% end %>
+  <h2><label for="block_encryption">SAML Assertion Encryption:</label></h2>
+  <p class="font-mono-xs margin-top-0" name="block_encryption"><%= service_provider.block_encryption %></p>
+<% end %>
+
+<h2>
+  <label for="saml_client_cert">Public certificate:</label>
```
aduth, line 76: It existed previously, though the `name` on the new `<details>` had caught my attention: Does the `for` attribute actually do anything for us? This doesn't seem like an appropriate use of a label, since it's not associated with an input.

https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Form_labelable

I could maybe see it as a case for [`aria-labelledby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) or [`aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute).

app/views/service_providers/_certificate_expiration_td.html.erb
```diff
@@ -1,14 +1,12 @@
...
-<% elsif cert.not_after < ServiceProviderCertificate.warning_period %>
-  <td style='color: black; background: yellow'>
-<% else %>
-  <td>
+<%
+ # locals: app
+ cert = app.certificate
+%>
+
+<%= content_tag(:td, class: cert.expiration_css_class) do %>
```
aduth, line 6: The new CSS classes have lower specificity than the inline styles, so the red background isn't being applied and the cell appears empty:

image

Because there's a background color being applied by `.usa-table td`:

https://github.com/uswds/uswds/blob/1908d1391bc59410624ca1934cc70b7404e8f443/src/stylesheets/core/placeholders/_table.scss#L88

We might need to do either...

- Apply the class to an inner element of the `td`
- Override the base `td` styles with greater specificity, like ...

```scss
.certificate-expired,
.usa-table td.certificate-expired {
  // ...
}
```

Unfortunately I don't see an option for the suggestion of my prior comment about using utility classes, so we may need the custom classes after all.

app/views/service_providers/show.html.erb
zachmargolis: Looks like they came from #366 from AxE matchers

app/views/service_providers/_certificate_expiration_td.html.erb
zachmargolis: Ooof, great catch. I went with the more specific styles in 119a9fed329bc328fd4d3f4c05529971e1586dbd

app/models/service_provider_certificate.rb
zachmargolis: Went with the more specific styles approach re: https://github.com/18F/identity-dashboard/pull/404/files#r604387482

app/views/service_providers/show.html.erb
zachmargolis: Ok no that was not right... looks like those have just been there a long time.... it's probably time for them to go, especially since this page isn't a form?

app/models/service_provider_certificate.rb
```diff
@@ -1,14 +1,30 @@
...
-class ServiceProviderCertificate < OpenSSL::X509::Certificate
-  def expiration_time_to_colorized_s
-    self.class.expiration_time_to_colorized_s(not_after)
+# A class to decorate a certificate, for easier warning and expiration
+class ServiceProviderCertificate
+  attr_reader :cert
+
+  # @param [OpenSSL::X509::Certificate,String] cert
+  def initialize(cert)
+    Rails.logger.info(cert.class)
```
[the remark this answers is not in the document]
zachmargolis: oh big whoops, that is my debugging 🙈 

thanks for catching!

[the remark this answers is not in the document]
zachmargolis: a7b31ccb7c8424cd50100c81bd8ef0937a81aa2d removed

app/views/service_providers/show.html.erb
orenyk: I wonder if they were copied over from the edit form?

orenyk (approved):
Looks good!

zachmargolis:
(overriding snyk because it's stalled out)

PR #405, opened 2021-03-31
Manage Multiple Certs (LG-2049)

zachmargolis:
(Corresponds to https://github.com/18F/identity-idp/pull/4851)

## Overview

Adds new functionality to support multiple certs.

In edit mode:
- Any existing cert can be deleted/removed via checkbox
- Only upload one cert at a time (optimizing for expected use cases)
- Adds some JS that scans the files uploaded to check for some common errors we know about:

## Screenies

Show VS Edit certificates section

| Show | Edit |
| --- | --- |
| Screen Shot 2021-03-30 at 8 43 01 PM | Screen Shot 2021-03-30 at 8 43 56 PM |

JS errors in various states:

| State | Screenshot |
| --- | --- |
| blank |  Screen Shot 2021-03-30 at 8 45 23 PM |
| private key uploaded | Screen Shot 2021-03-30 at 8 45 05 PM |
| unrecognized, includes DER files |  Screen Shot 2021-03-30 at 8 44 43 PM |


app/assets/stylesheets/application.css.scss.erb
```diff
@@ -110,3 +110,22 @@ $image-path: 'img';
...
+  }
+
+  dd {
+    margin: 0
+  }
+
+  dd::after {
+    content: '';
+    display: block;
+  }
```
aduth, line 130: Did you decide against inlining the `dt` and `dd`? We might not need this after all.

```suggestion
```

app/models/service_provider.rb
```diff
@@ -78,11 +78,11 @@ def redirect_uris=(uris)
...
 
-  # @return [ServiceProviderCertificate]
-  def certificate
-    @certificate ||= begin
-      if saml_client_cert
-        ServiceProviderCertificate.new(OpenSSL::X509::Certificate.new(saml_client_cert))
+  # @return [Array<ServiceProviderCertificate>]
+  def certificates
+    @certificates ||= [*certs, *saml_client_cert].map do |cert|
+      if cert
```
aduth, line 84: Could this ever be `nil`?

app/views/service_providers/_form.html.erb
```diff
@@ -112,10 +112,67 @@
                    hint: 'The name of the <a href="https://github.com/18F/identity-idp/tree/main/app/assets/images/sp-logos" target="_blank">logo image file</a> in the IdP. The login.gov team can add this image for you.'.html_safe %>
   <% end %>
 
-  <%= form.input :saml_client_cert,
-                 label: 'Public certificate',
-                 input_html: { class: 'usa-textarea' },
-                 hint: 'Your public certificate, which contains your public key, is needed for OpenID Connect (when using <i>private_key_jwt</i>) and for SAML. This must be <a href="https://en.wikipedia.org/wiki/Privacy-enhanced_Electronic_Mail" target="_blank">PEM encoded</a>, for example in the form: <pre><code>-----BEGIN CERTIFICATE-----<br>MIIDXTCCAkWgAwIBAgIJAJC1HiIAZAiIMA0GCSqGSIb3Df<br>BAYTAkFVMRMwEQYDVQQIDApTb21lLVN0YXRlMSEwHwYDVx<br>B7xxt8BVc69rHV15A0qyx77CLSj3tCx2IUXVqRs5mlvA==<br>-----END CERTIFICATE-----</code></pre>'.html_safe %>
+  <fieldset class="usa-fieldset">
+    <%= form.label 'Certificates', for: :saml_client_cert, class: 'usa-label' %>
```
aduth, line 116: This doesn't feel like a label anymore, since `saml_client_cert` input doesn't exist. Maybe `legend` could work, though if we're just going for the visual effect, a `div` could be fine to?

```suggestion
    <div class="usa-label">Certificates</div>
```

app/controllers/service_providers_controller.rb
```diff
@@ -140,6 +143,28 @@ def service_provider_params
   end
   # rubocop:enable MethodLength
 
+  def attach_cert
+    return if params.dig(:service_provider, :cert).blank?
+
+    x509 = OpenSSL::X509::Certificate.new(params[:service_provider].delete(:cert).read)
+
+    service_provider.certs ||= []
+    service_provider.certs << x509.to_pem
```
aduth, line 152: Do we care to do any duplicate prevention?

app/javascript/app/service-provider-form.js
```diff
@@ -76,4 +76,35 @@ $(function(){
       preview.textContent = logo_file.name;
     });
   }
+
+  /**
+   * @param {string,null} message
```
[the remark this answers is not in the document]
zachmargolis: oops, got my YARD and JSDoc mixed. Thanks!

app/models/service_provider.rb
zachmargolis: Oh, with the `*`... I guess not lol, great catch

aduth: > Oh, with the `*`... I guess not lol, great catch

I was originally going to comment about whether the array should be `[*certs, saml_client_cert]`, but I see the `nil` handling is quite handy 😄

app/controllers/service_providers_controller.rb
zachmargolis: Um, I don't feel too strongly. I guess my follow-up question would be, what kind of duplicate? Do you mean literal duplicate strings? Or duplicate PEM (ex line wrapping etc etc normalized)? Or duplicate attributes ex serial + issuer?

aduth: I'd only thought to mention it because I was able to upload the same cert file multiple times. It had some strange effect on how the "Remove" input worked, since we'd create them with identical IDs. But I suppose this wouldn't be likely to occur much in practice, and the effects weren't overly problematic anyways.

app/javascript/app/service-provider-form.js
```diff
@@ -76,4 +76,35 @@ $(function(){
...
+
+  var pemInput = document.querySelector('.js-pem-input');
+  if (pemInput) {
+    pemInput.addEventListener('change', function(event) {
+      var file = event.target.files[0];
+
+      document.querySelector('.js-pem-file-name').innerText = file ? file.name : null;
+
+      if (file) {
+        file.text().then(function(content) {
```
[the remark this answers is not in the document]
zachmargolis: I decided to make the server validation a little more thorough in f2401ec8c1a80a8c694287c1b75e417fc91bac93, and updated the JS to check for `.text` before calling it (makes the JS validation best-effort)

PR #408, opened 2021-04-01
Drop service_providers.saml_client_cert

zachmargolis:
This is OK to merge once https://github.com/18F/identity-dashboard/pull/407 has been deployed


spec/db/migrate/20210401171237_move_saml_client_cert_to_certs_spec.rb
```diff
@@ -1,35 +0,0 @@
-require 'rails_helper'
```
orenyk, line 1: **Question:** Is there a reason why you removed this spec file?

orenyk (approved):
One question, otherwise looks good!

spec/db/migrate/20210401171237_move_saml_client_cert_to_certs_spec.rb
zachmargolis: Because it's useless if the `saml_client_cert` column doesn't exist anymore. Either we'd have to stub it (ew) or find a way to roll the DB back to a particular migration just for these tests

orenyk: 🤦 that makes perfect sense, thanks!

orenyk: That's probably why you don't normally write migration tests 😛.

stevegsa (approved):

PR #412, opened 2021-04-06
[LG-4450] Add cert upload to feature spec for SP creation

orenyk:
Turned out to be a lot simpler than I thought 😄.

### DIFF
--- a/spec/features/service_providers_spec.rb
+++ b/spec/features/service_providers_spec.rb
@@ -24,7 +24,13 @@ feature 'Service Providers CRUD' do
 
       check 'email'
       check 'first_name'
-      click_on 'Create'
+
+      Tempfile.create(['cert', '.crt']) do |tmp|
+        File.write(tmp.path, build_pem)
+        attach_file('cert', tmp.path)
+
+        click_on 'Create'
+      end
 
       expect(page).to have_content('Success')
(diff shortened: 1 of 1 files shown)


spec/features/service_providers_spec.rb
```diff
@@ -21,6 +21,7 @@
       select user.teams[0].name, from: 'service_provider_group_id'
       select 'IAL2', from: 'Identity Assurance Level (IAL)'
       select 'AAL2', from: 'Authentication Assurance Level (AAL)'
+      attach_file 'cert', Rails.root.join('spec', 'fixtures', 'cert.crt')
```
zachmargolis, line 24: I added a `build_cert` helper to help generate valid PEMs, so if we wanted to write a tempfile rather than check these in, we could do something like:

```suggestion
      Tempfile.create(['cert', '.crt']) do |tmp|
        File.write(tmp.path, build_cert)
        attach_file 'cert', tmp.path
      end
```

zachmargolis (approved):
LGTM thanks for circling back!

spec/features/service_providers_spec.rb
orenyk: Ooh good call love it - thanks!

PR #413, opened 2021-04-08
Switch Dashboard to only push the latest cert

orenyk:


app/serializers/service_provider_serializer.rb
```diff
@@ -31,10 +32,18 @@ def agency_id
...
   end
 
-  def certs
-    object.certificates.map(&:to_pem)
+  # The one expiring latest - no way to know which was added last. Since this is
+  # temporary I think it's ok.
+  def cert
+    object.certificates.sort do |cert_a, cert_b|
+      cert_b.not_after - cert_a.not_after
+    end.first.to_pem
```
zachmargolis, line 40:
```suggestion
    object.certificates.sort_by(&:not_after).last
```

app/serializers/service_provider_serializer.rb
```diff
@@ -31,10 +32,18 @@ def agency_id
...
   end
 
+  # def certs
+  #   object.certificates.map(&:to_pem)
+  # end
```
zachmargolis, line 45: let's remove the comment for this, we can always get it back in source controller

zachmargolis (approved):
LGTM

app/serializers/service_provider_serializer.rb
orenyk: 🤦

orenyk: Oh I know what happened - I thought that I needed to call `to_pem` before I could call `not_after` so that's why I needed the block, but then when I figured that out I forgot to switch it. Thanks!

app/serializers/service_provider_serializer.rb
```diff
@@ -6,6 +6,7 @@ class ServiceProviderSerializer < ActiveModel::Serializer
     :assertion_consumer_logout_service_url,
     :attribute_bundle,
     :block_encryption,
+    :cert,
     :certs,
```
zachmargolis, line 10: remove the plural one here for now?

orenyk: d'oh!

PR #415, opened 2021-04-12
Accept RISC Security Event Tokens (LG-4413)

zachmargolis:
- Record them in the DB
- Show them in the UI

Here's a little tour!

* **Note**: I remembered to add the details/summary CSS fix after I took screenshots, so the triangle *is* there correctly

| notes | screenshot |
| --- | ---- |
| New nav item, view a table of security events for yourself | Screen Shot 2021-04-12 at 4 42 02 PM |
| mouse over the event name to see the full name | Screen Shot 2021-04-12 at 4 42 28 PM  | good ol' details/summary to see the full payload |  Screen Shot 2021-04-12 at 4 43 33 PM |
| pagination!! because there will be a lot of these |  Screen Shot 2021-04-12 at 4 41 51 PM |
| admins get a new dropdown, they can see events across all users | Screen Shot 2021-04-12 at 4 43 23 PM |

### DIFF
--- a/app/controllers/security_events_controller.rb
+++ b/app/controllers/security_events_controller.rb
@@ -3,8 +3,8 @@ class SecurityEventsController < ApplicationController
 
   def index
     @security_events = current_user.security_events.
-                                    order('issued_at DESC').
-                                    page(params[:page])
+                       order('issued_at DESC').
+                       page(params[:page])
 
     if @security_events.out_of_range?
       redirect_to security_events_path
@@ -15,8 +15,8 @@ class SecurityEventsController < ApplicationController
 
   def all
     @security_events = SecurityEvent.includes(:user).
-                                     order('issued_at DESC')
-                                     page(params[:page])
+                       order('issued_at DESC')
+                       page(params[:page])
 
     if @security_events.out_of_range?
       redirect_to security_events_all_path
--- a/app/controllers/security_events_controller.rb
+++ b/app/controllers/security_events_controller.rb
@@ -6,11 +6,7 @@ class SecurityEventsController < ApplicationController
                        order('issued_at DESC').
                        page(params[:page])
 
-    if @security_events.out_of_range?
-      redirect_to security_events_path
-    else
-      assign_pagination
-    end
+    assign_pagination
   end
 
   def all
@@ -18,11 +14,7 @@ class SecurityEventsController < ApplicationController
                        order('issued_at DESC').
                        page(params[:page])
 
-    if @security_events.out_of_range?
-      redirect_to security_events_all_path
-    else
-      assign_pagination
-    end
+    assign_pagination
   end
 
   private
--- a/app/controllers/security_events_controller.rb
+++ b/app/controllers/security_events_controller.rb
@@ -15,7 +15,7 @@ class SecurityEventsController < ApplicationController
 
   def all
     @security_events = SecurityEvent.includes(:user).
-                       order('issued_at DESC')
+                       order('issued_at DESC').
                        page(params[:page])
 
     if @security_events.out_of_range?
--- a/app/controllers/security_events_controller.rb
+++ b/app/controllers/security_events_controller.rb
@@ -1,0 +1,34 @@
+class SecurityEventsController < ApplicationController
+  before_action -> { authorize SecurityEvent }, only: %i[index all]
+
+  def index
+    @security_events = current_user.security_events
+      .order('issued_at DESC')
+      .page(params[:page])
+
+    if @security_events.out_of_range?
+      redirect_to security_events_path
+    else
+      assign_pagination
+    end
+  end
+
+  def all
+    @security_events = SecurityEvent.includes(:user)
+      .order('issued_at DESC')
+      .page(params[:page])
+
+    if @security_events.out_of_range?
+      redirect_to security_events_all_path
+    else
+      assign_pagination
+    end
+  end
+
+  private
+
+  def assign_pagination
+    @prev_page = @security_events.prev_page && url_for(page: @security_events.prev_page)
+    @next_page = @security_events.next_page && url_for(page: @security_events.next_page)
+  end
+end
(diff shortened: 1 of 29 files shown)


app/views/security_events/_table.html.erb
```diff
@@ -0,0 +1,84 @@
...
+            <td>
+              <%= l(security_event.issued_at, format: :long) %>
+            </td>
+            <td>
+              <% if security_event.raw_event %>
+                <details>
+                  <summary class="summary-fix">Show Payload</summary>
+                  <pre><code><%= JSON.pretty_generate(JSON.parse(security_event.raw_event)) %></code></pre>
+                </details>
+              <% end %>
```
zachmargolis, line 45: Question maybe for @aduth, right now when the summary expands, it blows up the size of the entire table because it changes the column widths. Do you have any other ideas for showing the full details? I guess I could do pairs of rows and have a `colspan` kinda thing going so that it expands downwards only?

app/controllers/security_events_controller.rb
```diff
@@ -0,0 +1,34 @@
+class SecurityEventsController < ApplicationController
+  before_action -> { authorize SecurityEvent }, only: %i[index all]
+
+  def index
+    @security_events = current_user.security_events.
+                       order('issued_at DESC').
+                       page(params[:page])
+
+    if @security_events.out_of_range?
```
aduth, line 9: When I have no security events, it appears that `#out_of_range?` is `true` (somewhat surprisingly), so this triggers an infinite redirect loop.

Maybe...

```suggestion
    if !@security_events.empty? && @security_events.out_of_range?
```

app/controllers/security_events_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+      assign_pagination
+    end
+  end
+
+  def all
+    @security_events = SecurityEvent.includes(:user).
+                       order('issued_at DESC').
+                       page(params[:page])
+
+    if @security_events.out_of_range?
```
aduth, line 21: Same note as above.

app/views/security_events/_table.html.erb
aduth: > Question maybe for @aduth, right now when the summary expands, it blows up the size of the entire table because it changes the column widths. Do you have any other ideas for showing the full details? I guess I could do pairs of rows and have a `colspan` kinda thing going so that it expands downwards only?

That could work, yeah. Maybe paired with an expanding control, like in [this example](https://css-tricks.com/table-with-expando-rows/).

I think a simpler alternative could be to create a dedicated detail page for each event, and include the payload on that.

app/forms/security_event_form.rb
```diff
@@ -0,0 +1,78 @@
...
+        next
+      end
+
+      return payload if payload
+
+      errors.add(:jwt, 'could not verify JWT with any known keys')
+      {}
+    end
+  end
+  # rubocop:enable Metrics/MethodLength
```
mitchellhenke, line 56: I vote to remove or increase this limit

app/policies/security_event_policy.rb
```diff
@@ -0,0 +1,16 @@
+class SecurityEventPolicy < BasePolicy
+  attr_reader :current_user, :model
+
+  def initialize(current_user, model)
+    @current_user = current_user
+    @model = model
+  end
+
+  def index?
+    true
```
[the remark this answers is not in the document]
zachmargolis: whoops, great catch!

spec/policies/security_event_policy_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  subject(:policy) { SecurityEventPolicy.new(user, SecurityEvent) }
+
+  describe '#index?' do
+    context 'for a non-admin user' do
+      let(:user) { build(:user) }
+
+      it 'is true' do
+        expect(policy.index?).to eq(true)
+      end
+    end
```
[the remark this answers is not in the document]
zachmargolis: thanks!

app/controllers/security_events_controller.rb
zachmargolis: giant facepalm... but I'm worried that a page too far out ex (1000) would be both empty and out of range... will have to noodle on this, maybe just easier to just have the page be empty

app/forms/security_event_form.rb
zachmargolis: This codebase still uses CodeClimate to lint... and the `.rubocop.yml` is so far out of date I can't even run rubocop locally... unsure where to change that limit 😬

app/views/security_events/_table.html.erb
zachmargolis: Good point, will go with the detail page approach

zachmargolis: Added details page in 5a31759d9575acc5f3ea99eb2d23254de16fc1ee

Screen Shot 2021-04-13 at 9 08 24 AM

mitchellhenke (approved):
👍🏼 👍🏼

app/forms/security_event_form.rb
mitchellhenke: I will take a look into this

app/forms/security_event_form.rb
```diff
@@ -0,0 +1,78 @@
...
+  def subject
+    payload.dig('events', event_type, 'subject') || {}
+  end
+
+  def user
+    subject_type = subject['subject_type']
+
+    case subject['subject_type']
+    when 'email'
+      User.find_by(email: subject['email'])
```
stevegsa, line 71: you're very close to 100% coverage :-)

app/services/idp_public_keys.rb
```diff
@@ -0,0 +1,35 @@
+# Ideally this could share an implementation with OmniAuth::LoginDotGov::IdpConfiguration
+# However at this time, that class only exposes one public key
+class IdpPublicKeys
+  # @return [Array<OpenSSL::PKey::PKey>]
+  def self.all
+    @all ||= new.load_all
```
stevegsa, line 6: you're very close to 100% coverage :-)

PR #416, opened 2021-04-13
Fix security_events/all routing

zachmargolis:
:giant_facepalm:

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -22,8 +22,8 @@ Rails.application.routes.draw do
   get '/service_providers/all' => 'service_providers#all'
   resources :service_providers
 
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -22,8 +22,8 @@
   get '/service_providers/all' => 'service_providers#all'
   resources :service_providers
 
-  resources :security_events, only: %i[index show]
   get '/security_events/all' => 'security_events#all'
+  resources :security_events, only: %i[index show]
```
zachmargolis, line 26: when the order is backwards it goes to `#show`, `id: 'all'`

mitchellhenke (approved):

PR #417, opened 2021-04-13
Run rubocop on Circle CI and update rubocop rules

mitchellhenke:
Similar to https://github.com/18F/identity-idp/pull/4055

CodeClimate runs a different version and set of rules for Rubocop than we have in our Gemfile and `.rubocop.yml`. This PR moves the rubocop check into CircleCI so running rubocop locally and the checks in CI are the same. Rules from the IDP were brought over to more closely match. I also removed slim since it wasn't being used.


.codeclimate.yml
```diff
@@ -62,9 +62,6 @@ plugins:
       - XXX
   pep8:
     enabled: true
-  rubocop:
```
zachmargolis, line 65: maybe we just remove all codeclimate like we did in the IDP?

.circleci/config.yml
```diff
@@ -61,7 +61,7 @@ jobs:
 
             make test
             yarn test
-            bundle exec slim-lint app/views
```
aduth, line 64: Should we remove `.slim-lint.yml` ?

.rubocop.yml
```diff
@@ -22,14 +20,19 @@ AllCops:
...
+    - 'public/**/*'
+    - 'node_modules/**/*'
   TargetRubyVersion: 2.5
   TargetRailsVersion: 6.1
   UseCache: true
+  DisabledByDefault: true
 
 Rails:
-  Enabled: true
+  Enabled: false
```
aduth, line 33: Curious if I'm following correctly. If we disable rules by default, do we need to have any of these where we explicitly set `Enabled: false` ?

aduth (approved):
Nice 👍

zachmargolis (approved):
LGTM! love this

.circleci/config.yml
mitchellhenke: Yep, thank you

.rubocop.yml
mitchellhenke: We do not, will remove

.codeclimate.yml
mitchellhenke: I think we still have code climate in the IDP? 
https://github.com/18F/identity-idp/blob/main/.codeclimate.yml

zachmargolis:
LOL @ code coverage decreasing... LMK if you need an admin merge

mitchellhenke:
@zachmargolis code climate will not beat me this time https://github.com/18F/identity-dashboard/pull/417/commits/ff4e4b6813801124a97a49938d0a044aba7789e6

PR #420, opened 2021-04-15
RISC UX Improvements: filter/search by user email (LG-4414)

zachmargolis:
Adds new search field

| notes | screenshot |
| --- |--- |
| search field, empty | Screen Shot 2021-04-15 at 9 59 31 AM |
| filters results by email | Screen Shot 2021-04-15 at 9 59 48 AM |
| flashes a warning for unknown emails |  Screen Shot 2021-04-15 at 10 00 05 AM |

### DIFF
--- a/app/controllers/security_events_controller.rb
+++ b/app/controllers/security_events_controller.rb
@@ -1,5 +1,5 @@
 class SecurityEventsController < ApplicationController
-  before_action -> { authorize SecurityEvent }, only: %i[index all]
+  before_action -> { authorize SecurityEvent }, only: %i[index all search]
   before_action -> { authorize security_event }, only: %i[show]
 
   rescue_from ActiveRecord::RecordNotFound do
@@ -15,7 +15,13 @@ class SecurityEventsController < ApplicationController
   end
 
   def all
-    @security_events = SecurityEvent.includes(:user).
+    scope = SecurityEvent.includes(:user)
+
+    if params[:user_uuid].present? && (@user = User.find_by(uuid: params[:user_uuid]))
+      scope = scope.where(user_id: @user.id)
+    end
+
+    @security_events = scope.
                        order('issued_at DESC').
                        page(params[:page])
 
@@ -24,6 +30,21 @@ class SecurityEventsController < ApplicationController
 
   def show
     @security_event = security_event
+  end
+
+  def search
+    email = params[:email]
+
+    if email.present?
+      if (user = User.find_by(email: email))
+        redirect_to security_events_all_path(user_uuid: user.uuid)
+        return
+      else
+        flash[:warning] = "Could not find a user with email #{email}"
+      end
+    end
+
+    redirect_to security_events_all_path
   end
 
   private
--- a/app/views/security_events/all.html.erb
+++ b/app/views/security_events/all.html.erb
@@ -1,5 +1,18 @@
 <h2>All Security Events</h2>
 
+<%= form_with url: security_events_search_path, method: :post do |form| %>
+  <fieldset class="usa-fieldset">
+    <legend class="usa-legend">
+      Filter Events By
+    </legend>
+
+    <%= form.label :email, 'Email', class: 'usa-label' %>
+    <%= form.text_field :email, class: 'usa-input', value: @user&.email %>
+
+    <%= form.submit 'Search', class: 'usa-button margin-top-1' %>
+  </fieldset>
+<% end %>
+
 <%= render partial: 'table',
            locals: {
              security_events: @security_events,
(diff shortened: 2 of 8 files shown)


app/controllers/security_events_controller.rb
```diff
@@ -26,6 +32,21 @@ def show
     @security_event = security_event
   end
 
+  def search
+    email = params[:email]
+
+    if email.present?
+      if (user = User.find_by(email: email))
+        redirect_to security_events_all_path(user_uuid: user.uuid)
```
zachmargolis, line 40: My feeling is that emails don't belong in URLs, so that's why the this gets turned into a UUID

zachmargolis: Also that GET url params are easier to work with for a search, instead of doing POST body to accommodate an email

app/views/security_events/all.html.erb
```diff
@@ -1,5 +1,18 @@
 <h2>All Security Events</h2>
 
+<%= form_with url: security_events_search_path, method: :post do |form| %>
+  <fieldset class="usa-fieldset">
+    <legend class="usa-legend">
+      Filter Events By
+    </legend>
+
+    <%= form.label :email, 'Email', class: 'usa-label' %>
```
juliaelman, line 9: s Email/Search by email address

app/controllers/security_events_controller.rb
mitchellhenke: Not major, but any reason we couldn't do this in the existing `all` route?

juliaelman (changes requested):
One small change. Also! Can we use the same pattern we use on login.gov (see screenshot below)?

Screen Shot 2021-04-15 at 1 28 10 PM

app/controllers/security_events_controller.rb
zachmargolis: you mean `POST /security_events/all => security_events#search`? Yeah that makes sense

mitchellhenke: Sorry, I mean something like `GET /security_events/all?email=mitchell.henke@gsa.gov` redirects to `GET /security_events/all?user_uuid=abcd1234`. Should we need to support more field searching in the future, it becomes along the lines of:

```ruby
def all
  email = params.delete(:email)
  if email.present?
    user = User.find_by(uuid: params[:user_uuid])
    params[:user_uuid] = user&.uuid
    redirect_to security_events_all_path(params) and return
  end
# ...
end
```

The flash could still be added in there somewhere.

zachmargolis: If we did a GET like that, the email address from the URL is still in the logs, which is what I was trying to avoid

mitchellhenke: ahhh, got it

app/views/security_events/all.html.erb
zachmargolis: Screen Shot 2021-04-15 at 11 03 08 AM

@juliaelman updated in f6ca637, I think that covers it?

juliaelman (approved):

app/controllers/security_events_controller.rb
```diff
@@ -26,6 +32,21 @@ def show
     @security_event = security_event
   end
 
+  def search
+    email = params[:email]
+
+    if email.present?
+      if (user = User.find_by(email: email))
```
stevegsa, line 39: Do we want to confine this to the groups the user is in?

zachmargolis: the "all" page is for admins only, so I think it's fine to show across all groups

app/assets/stylesheets/application.css.scss.erb
```diff
@@ -146,3 +146,11 @@ $image-path: 'img';
     display: block;
   }
 }
+
+// Search overrides based on identity-site
+
+.usa-search-input--joined-right {
+  border-top-right-radius: 0;
+  border-bottom-right-radius: 0;
```
aduth, line 154: I hadn't noticed this before about the [search component](https://design.login.gov/components/search/), but it should ideally be fixed in the design system as well. It looks like they're intended to be squared in USWDS ([source](https://github.com/uswds/uswds/blob/0f1f4d0a12bb3981a52cc882c78b79931e6b004c/src/stylesheets/components/_search.scss#L82-L84)), but it's inadvertently overridden in our design system input styles ([source](https://github.com/18F/identity-style-guide/blob/ba2b1f7863a59a1fcbd47b713cfb7a4455813ab2/src/scss/components/_inputs.scss#L10)).

I can make a note to follow-up on this.

Ideally, we shouldn't need to have this class at all, since it should be the behavior of the default search input placed within a grid container of whatever size we prefer.

aduth (approved):
:+1:

app/assets/stylesheets/application.css.scss.erb
zachmargolis: Thanks!

PR #422, opened 2021-04-16
[LG-4224] Add PaperTrail versioning to Users, Teams, and UserTeams

orenyk:
Resolves LG-4224

Also add UserTeam factory and model specs for all three models


achapm (approved):
LGTM

PR #423, opened 2021-04-16
[LG-4072] Remove middle_name from attribute bundle input

orenyk:
Resolves LG-4072

Add ServiceProviderHelper#sp_attribute_bundle to filter out invalid attributes from the show view


achapm (approved):
LGTM

PR #425, opened 2021-04-19
Fix the setup:script

jmhooper:
**Why**: So the `make setup` works as expected

**How**: The dev:prime task was having issues with a foreign key constraint on the agencies table. This commit avoids that by using an agency that is added by the agency seeder.


zachmargolis (approved):
LGTM

PR #427, opened 2021-04-20
[LG-4224] Ensure that PaperTrail tracks user management events

orenyk:
Resolves LG-4224

Even though the join records were being destroyed when removing a user
from a team, PaperTrail wasn't being triggered and no version was being
saved. This commit adds `dependent: :destroy` to the `has_many :users`
association for Teams, which properly triggers the PaperTrail callback.

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -7,9 +7,8 @@ class Team < ApplicationRecord
 
   has_many :service_providers, dependent: :nullify, foreign_key: 'group_id',
                                inverse_of: :team
-  has_many :user_teams, dependent: :nullify, foreign_key: 'group_id',
-                        inverse_of: :team
-  has_many :users, through: :user_teams
+  has_many :user_teams, foreign_key: 'group_id', inverse_of: :team
+  has_many :users, dependent: :destroy, through: :user_teams
 
   validates :name, presence: true, uniqueness: true
 
--- a/spec/features/manage_users_spec.rb
+++ b/spec/features/manage_users_spec.rb
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 feature 'manage users', :js do
-  scenario 'adding and removing users by email address' do
+  scenario 'adding and removing users by email address', versioning: true do
     team = create(:team)
     user = create(:user, teams: [team])
     user_to_remove = create(:user, teams: [team])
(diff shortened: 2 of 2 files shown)


app/forms/manage_users_form.rb
```diff
@@ -14,28 +14,34 @@ def initialize(team)
...
-    result = team.update(users: users_from_user_emails)
-    errors.merge!(team.errors)
-    result
+    ActiveRecord::Base.transaction do
+      destroy_removed_user_teams
+      add_new_users
+    end
+    true
+  rescue ActiveRecord::ActiveRecordError => e
+    errors.add(:base, "Unable to update users: #{e.message}")
```
orenyk, line 28: This may not play nicely with all `ActiveRecordError` children but it also should be _very_ unlikely. I can remove the inclusion of `e.message` to keep it simpler.

app/models/user.rb
```diff
@@ -4,7 +4,7 @@ class User < ApplicationRecord
   has_paper_trail on: %i[create update destroy]
 
   devise :trackable, :timeoutable
-  has_many :user_teams, dependent: :nullify
+  has_many :user_teams, dependent: :destroy
```
achapm, line 7: This change is not needed

achapm (approved):
See note above, otherwise LGTM

app/forms/manage_users_form.rb
```diff
@@ -14,28 +14,34 @@ def initialize(team)
...
+  attr_reader :emails_to_remove, :emails_to_add
 
-  def missing_emails
-    existing_emails = existing_users.map(&:email)
-    user_emails.filter { |email| !existing_emails.include? email }
+  def destroy_removed_user_teams
+    users_to_remove = User.where(email: emails_to_remove).pluck(:id)
+    UserTeam.where(group_id: team.id, user_id: users_to_remove).each do |ut|
+      ut.destroy! # we want to trigger the paper_trail callback
+    end
```
zachmargolis, line 40: does the `has_many :through` do this? Is there any way we can get this desired behavior through `dependent: :destroy` annotations?

orenyk: I tried - it didn't work unfortunately. The records are actually being destroyed but PaperTrail wasn't being triggered.

I'm 99% sure I tested this with the feature spec before I started messing with the Form Object but I'm going to check one more time.

orenyk: @zachmargolis turns out I forgot to test this again after I remembered to manually enable PaperTrail for the feature spec - that indeed works when applied to the `has_many :users` association. Thanks so much!

app/forms/manage_users_form.rb
orenyk: No longer necessary

zachmargolis (approved):
LGTM

PR #428, opened 2021-04-20
Cleanup the dashboard application.yml

jmhooper:
This commit restructures the application.yml.example file so it's format matches the other repositories. This change is the first step in migrating the dashboard to IdentityConfig and using Hostdata to manage configs here.


zachmargolis (approved):
LGTM

PR #429, opened 2021-04-20
Stop using env vars to configure rack timeout

jmhooper:
**Why**: So that the configuration for rack-timeout matches the configuration in the IDP and in PKI


zachmargolis (approved):
LGTM

PR #430, opened 2021-04-21
Upgrade Hostdata and start using IdentityConfig

jmhooper:
**Why**: So that the way configs are delivered to this app matches the other apps


jmhooper:
Originally my plan was to slim these up by having separate PRs for updating Hostdata and then adding IdentityConfig. I ended up spending too much time trying to drag the active script and spec forward so I decided to integrate IdentityConfig in place along with the upgrade. That allowed my to simply remove the activate code.

Gemfile
```diff
@@ -12,9 +12,8 @@ gem 'colorize'
 gem 'devise', '~> 4.7.3'
 gem 'dotenv-rails', '~> 2.7', '>= 2.7.6'
 gem 'enum_help'
-gem 'figaro'
 gem 'httparty'
-gem 'identity-hostdata', git: 'https://github.com/18F/identity-hostdata.git', tag: 'v1.0.1'
+gem 'identity-hostdata', git: 'https://github.com/18F/identity-hostdata.git', tag: 'v3.1.0'
```
jmhooper, line 16: I'll need to publish a new tag with [this change](https://github.com/18F/identity-hostdata/pull/27) and add it here before merge

lib/identity_config.rb
```diff
@@ -0,0 +1,80 @@
...
+    config.add(:aws_region, type: :string)
+    config.add(:aws_logo_bucket, type: :string)
+    config.add(:db_pool, type: :integer)
+    config.add(:dp_reaping_frequency, type: :integer)
+    config.add(:certificate_expiration_warning_period, type: :integer)
+    config.add(:dashboard_api_token, type: :string)
+    config.add(:email_recipients, type: :string)
+    config.add(:idp_sp_url, type: :string)
+    config.add(:idp_url, type: :string)
+    config.add(:logo_upload_enabled, type: :boolean)
```
zachmargolis, line 59: I'd really like us to move this class into hostdata soon, and then make this long `build_store` more DSL-y and move it into `application.rb` or some other initializer

zachmargolis (approved):
LGTM

lib/identity_config.rb
jmhooper: Agreed, I was thinking through that while putting this together.

PR #433, opened 2021-04-21
Fix bin/activate and make sure it works

jmhooper:
**Why**: My last commit had a bug where we were trying to run bash in a ruby script


zachmargolis (approved):
LGTM

PR #437, opened 2021-05-06
[LG-3742] Add notice about IAA requirement to SP pages

orenyk:
Resolves LG-3742


app/views/layouts/_messages.html.erb
```diff
@@ -1,12 +1,14 @@
...
-    <% if msg.is_a?(String) %>
-      <% if name == 'notice' then name = 'info' end %>
-      <div class="usa-alert usa-alert__paragraph usa-alert--<%= name %>" >
-        <div class="usa-alert__body">
-          <p class="usa-alert__text"><%= msg %></p>
+  <div class="margin-bottom-2">
+    <% flash.each do |name, msg| %>
+      <% if msg.is_a?(String) %>
+        <% if name == 'notice' then name = 'info' end %>
+        <div class="usa-alert usa-alert__paragraph usa-alert--<%= name %>" >
```
zachmargolis, line 6: `if ... then` is super uncommon, WDYT of adding a new variable and using a ternary?

```suggestion
        <% alert_name = name == 'notice' ? 'info' : name %>
        <div class="usa-alert usa-alert__paragraph usa-alert--<%= alert_name %>" >
```

config/locales/en.yml
```diff
@@ -68,6 +68,7 @@ en:
     confirm_delete: Are you sure you want to delete this user?
     service_provider_deleted: Success! You have deleted %{issuer}
     service_provider_saved: Success! You have saved %{issuer}
+    service_provider_iaa_notice: Please note that you will need a signed Inter-Agency Agreement with login.gov in order to go live in production. Visit <a href="https://developers.login.gov/production/" target="_blank">https://developers.login.gov/production</a> for more details or email <a href="mailto:partners@login.gov">partners@login.gov</a> with any questions.
```
zachmargolis, line 71: If you suffix this key with `_html`, it will automatically be marked as HTMLsafe by the i18n gem

app/views/layouts/_messages.html.erb
```diff
@@ -1,12 +1,14 @@
...
+          <div class="usa-alert__body">
+            <p class="usa-alert__text"><%= sanitize(msg) %></p>
```
zachmargolis, line 8: Do we need to explicitly allow `<a>` tags in `sanitize`? Or does this go away if we suffix the `_html` key correctly? (sometimes serializing things to a flash loses that data...)

config/locales/en.yml
orenyk: oooh neat trick!

app/views/layouts/_messages.html.erb
orenyk: Just indented this partial to add some padding but I'm happy to update this.

app/views/layouts/_messages.html.erb
orenyk: We'll still need `sanitize`, I think, but it's working as expected with no other modifications (tested both locally and in my personal sandbox).

config/locales/en.yml
orenyk: This didn't work 😕 - I think we need to add `sanitize` (based on [this thread](https://stackoverflow.com/questions/17375232/rails-i18n-html-suffix-rule-and-translate-helper-called-from-a-controller) - it's old but makes sense since we're calling it in the controller).

zachmargolis (approved):
LGTM

orenyk:
@zachmargolis good call! I'll get some feature specs in there to make sure it's rendering correctly on the right pages.

achapm (approved):
LGTM

PR #444, opened 2021-05-25
Manage EventBridge API destinations for ServiceProviders (LG-4532)

zachmargolis:
(this just an outline, it's in draft mode)

### DIFF
--- a/app/services/risc_destination_updater.rb
+++ b/app/services/risc_destination_updater.rb
@@ -1,0 +1,53 @@
+require 'aws-sdk-eventbridge'
+
+class RiscDestinationUpdater
+  attr_reader :service_provider
+
+  def initialize(service_provider)
+    @service_provider = service_provider
+  end
+
+  def update!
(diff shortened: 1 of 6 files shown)


app/services/risc_destination_updater.rb
```diff
@@ -0,0 +1,53 @@
...
+      connection_arn: connection_arn,
+      description: "Destination for #{service_provider.friendly_name}",
+      invocation_endpoint: service_provider.push_notification_url,
+      http_method: 'POST',
+    }
+  end
+
+  def connection_arn
+    eventbridge_client.list_connections(
+      name_prefix: '???', # TODO: fixme
```
zachmargolis, line 45: cc @akrito unsure how we're naming these connections, looking for some pointers here

zachmargolis: sounds good to me! will update it

mitchellhenke (approved):

PR #450, opened 2021-06-03
[LG-4680] Add state and local agencies

orenyk:
Resolves LG-4680

Data sourced from https://en.wikipedia.org/wiki/List_of_states_and_territories_of_the_United_States and we already had DC in there.

### DIFF
--- a/config/agencies.yml
+++ b/config/agencies.yml
@@ -287,4 +287,112 @@
 149:
   name: Vietnam Education Foundation
 21:
-  name: 'State, local governments or tribes'
+  name: Other local or tribal government
+150:
+  name: State of Alabama
+151:
+  name: State of Alaska
+152:
+  name: State of Arizona
+153:
+  name: State of Arkansas
+154:
+  name: State of California
+155:
+  name: State of Colorado
+156:
+  name: State of Connecticut
+157:
+  name: State of Delaware
+158:
+  name: State of Florida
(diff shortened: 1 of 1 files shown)


achapm (approved):
LGTM

config/agencies.yml
```diff
@@ -288,3 +288,111 @@
   name: Vietnam Education Foundation
 21:
   name: 'State, local governments or tribes'
```
zachmargolis, line 290: should we deprecate this one or rename it to "other local or tribal government" ?

zachmargolis (approved):
LGTM... too bad we have to update this in multiple places now 😭

config/agencies.yml
orenyk: Good call, I thought about it and then forgot after all the typing 😄. I think we should rename since we don't really have a mechanism to "deprecate" or mark an agency as inactive, so I think I'll update it to "Other local or tribal government" (although most local governments should fall under their respective states). Thanks!

zachmargolis: Other local things could be like city or county? maybe? 🤷

orenyk: Well if the goal of the Agency is to aggregate UUIDs within a given scope I think the state seems like the appropriate level for "things that aren't federal", IMO.

PR #451, opened 2021-06-04
Update RiscDestinationUpdater to handle more AWS object graph

zachmargolis:
Here's my mental model of what I'm trying to create (the boxes in blue)

RISC EventBridge Architecture (1)

I went a lot lighter on the test coverage this time because at the end of the day it's all stubbed, and more tests won't catch misconfigurations

### DIFF
--- a/app/services/risc_destination_updater.rb
+++ b/app/services/risc_destination_updater.rb
@@ -9,18 +9,123 @@ class RiscDestinationUpdater
 
   def update
     if service_provider.push_notification_url.present?
-      if api_destination_exists?
-        eventbridge_client.update_api_destination(api_destination_attributes)
-      else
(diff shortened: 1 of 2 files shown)


mitchellhenke (approved):

app/services/risc_destination_updater.rb
```diff
@@ -9,18 +9,123 @@ def initialize(service_provider)
...
+  end
+
+  def put_targets(api_destination_arn)
+    eventbridge_client.put_targets(
+      rule: rule_name,
+      event_bus_name: event_bus_name,
+      targets: [
+        id: target_id,
+        arn: api_destination_arn,
+        input_path: '$.detail',
```
zachmargolis, line 76: this will also need a new `role_arn` key where the IAM role will go cc @akrito

PR #452, opened 2021-06-09
add paper trail delete reports to teams and users

achapm:
LG-4520
Add query object for looking up deleted UserTeams for a given team

ACs:
- Put together a query object to easily look up UserTeam deletions (e.g. UserTeam.deleted_versions)
- Add a method on team model to report all deleted users and dates


app/models/user.rb
```diff
@@ -27,6 +27,34 @@ def scoped_service_providers
...
+
+  def user_deletion_report_item(record)
+    {
+      user_id: record[0]['user_id'],
+      user_email: User.find_by(id: record[0]['user_id'])&.email,
+      team_id: record[0]['group_id'],
+      team_name: Team.find_by(id: record[0]['group_id'])&.name,
+      removed_at: record[1],
+      whodunnit_id: record[2],
+      whodunnit_email: User.find_by(id: record[2])&.email
```
zachmargolis, line 44: can we destructure `record` to make it clearer what the fields are? I can't tell what the `record[0]` is so we need a better name there obvi

```suggestion
  def user_deletion_report_item(record)
    something, removed_at, whodunnit_id = record
    {
      user_id: something['user_id'],
      user_email: User.find_by(id: something['user_id'])&.email,
      team_id: something['group_id'],
      team_name: Team.find_by(id: something['group_id'])&.name,
      removed_at: removed_at,
      whodunnit_id: whodunnit_id,
      whodunnit_email: User.find_by(id: whodunnit_id)&.email
```

app/models/user.rb
```diff
@@ -27,6 +27,34 @@ def scoped_service_providers
...
+  end
+
+  def user_deletion_history_report
+    user_deletion_history
+      .order(created_at: :desc)
+      .limit(5000)
+      .pluck(:object, :created_at, :whodunnit)
+      .map { |record|
+        user_deletion_report_item(record)
+      }
```
zachmargolis, line 55: I hope rubocop will get mad at this, but if not, I will.... I prefer braces syntax for blocks on the same line and `do...end` for multiline blocks so this should be one of two things:


```suggestion
      .map { |record| user_deletion_report_item(record) }
```

```suggestion
      .map do |record|
        user_deletion_report_item(record)
      end
```

spec/support/paper_trail_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+  was_enabled_for_request = PaperTrail.request.enabled?
+  PaperTrail.enabled = true
+  PaperTrail.request.enabled = true
+  begin
+    yield
+  ensure
+    PaperTrail.enabled = was_enabled
+    PaperTrail.request.enabled = was_enabled_for_request
+  end
+end
```
zachmargolis, line 13: we can simplify this using the implict begin in a method
```suggestion
  yield
ensure
  PaperTrail.enabled = was_enabled
  PaperTrail.request.enabled = was_enabled_for_request
end
```

app/models/team.rb
```diff
@@ -18,6 +18,37 @@ def to_s
...
+  end
+
+  def user_deletion_history_report(email = nil)
+    user_deletion_history.
+      order(created_at: :desc).
+      limit(5000).
+      pluck(:object, :created_at, :whodunnit).
+      select { |record|
+        email.nil? || User.find_by(id: record[0]['user_id'])&.email == email
+      }.
```
orenyk, line 46: **suggestion:** You can probably use argument destructuring here as well to keep this more readable:

```suggestion
      select { |object, _, _|
        email.nil? || User.find_by(id: object['user_id'])&.email == email
      }.
```

config/environments/test.rb
```diff
@@ -54,4 +54,9 @@
   config.active_storage.service = :test
   # Allow SVG's only because we will always serve them in an <img> element
   config.active_storage.content_types_to_serve_as_binary -= ['image/svg+xml']
+
+  # disable PaperTrail for the entire ruby process.
```
orenyk, line 58: **question:** Are you sure this is necessary? I think it should be [disabled by default](https://github.com/paper-trail-gem/paper_trail#7b-rspec).

spec/support/paper_trail_helper.rb
```diff
@@ -0,0 +1,13 @@
+# enable PaperTrail for specific tests - you can add a with_versioning test helper method
```
orenyk, line 1: **question:** again, is this necessary? I think `with_versioning` is built in to `paper_trail` and you can also add `versioning: true` to the test definition, i.e. https://github.com/18F/identity-dashboard/blob/main/spec/models/user_team_spec.rb#L4.

orenyk (approved):
Looks good overall, I agree with @zachmargolis's comments about destructuring and I'm not sure we need all of the test configuration stuff, so please check if it's needed. Nice work!!

app/models/user.rb
```diff
@@ -27,6 +27,36 @@ def scoped_service_providers
...
+      removed_at: deleted_record['removed_at'],
+      whodunnit_id: deleted_record['whodunnit_id'],
+      whodunnit_email: User.find_by(id: deleted_record['whodunnit_id'])&.email,
+    }
+  end
+
+  def user_deletion_history_report
+    user_deletion_history.
+      order(created_at: :desc).
+      limit(5000).
```
zachmargolis, line 51: let's make limit an optional argument

```suggestion
  def user_deletion_history_report(limit: 5000)
    user_deletion_history.
      order(created_at: :desc).
      limit(limit).
```

app/models/team.rb
```diff
@@ -18,6 +18,39 @@ def to_s
...
+      removed_at: deleted_record['removed_at'],
+      whodunnit_id: deleted_record['whodunnit_id'],
+      whodunnit_email: User.find_by(id: deleted_record['whodunnit_id'])&.email,
+    }
+  end
+
+  def user_deletion_history_report(email = nil)
+    user_deletion_history.
+      order(created_at: :desc).
+      limit(5000).
```
zachmargolis, line 42: let's make limit an option here too

```suggestion
  def user_deletion_history_report(email: nil, limit: 5000)
    user_deletion_history.
      order(created_at: :desc).
      limit(limit).
```

zachmargolis (approved):
LGTM

PR #454, opened 2021-06-15
[LG-4446] Add PKCE option to service provider form

orenyk:
Resolves LG-4446

* Adds additional enumerated option for protocol and updates the generic
  OIDC option to specify private_key_jwt.
* Update JS to appropriately hide the certificate fields when OIDC PKCE
  is selected
* Add necessary acronyms to ActiveSupport::Inflector

### DIFF
--- a/app/javascript/app/service-provider-form.js
+++ b/app/javascript/app/service-provider-form.js
@@ -10,7 +10,11 @@ $(function(){
 
   var toggle_form_fields = function(id_protocol){
     switch(id_protocol){
-      case 'openid_connect':
+      case 'openid_connect_private_key_jwt':
+        $('.saml-fields').hide();
+        $('.oidc-fields').show();
+        break;
+      case 'openid_connect_pkce':
         $('.saml-fields').hide();
         $('.oidc-fields').show();
         break;
@@ -24,39 +28,14 @@ $(function(){
     }
   }
 
-  var issuer_protocol_name = function(id_protocol){
-    switch(id_protocol){
-      case 'openid_connect':
-        return 'openidconnect';
-      case 'saml':
-        return 'SAML:2.0';
-      default:
-        return "";
-    }
-  }
-
   var id_protocol = function(){
     return $('input[name="service_provider[identity_protocol]"]:checked').val();
-  }
-
-  var update_issuer = function() {
-    var protocol = issuer_protocol_name(id_protocol());
-    var department = $('#service_provider_issuer_department').val();
-    var app = $('#service_provider_issuer_app').val();
-
-    var issuer_string = SERVICE_PROVIDER_ISSUER_TEMPLATE
-      .replace('{protocol}', protocol)
-      .replace('{department}', department)
-      .replace('{app}', app);
-
-    $('#service_provider_issuer').val(issuer_string);
   }
 
   toggle_form_fields(id_protocol());
 
   $('input[name="service_provider[identity_protocol]"]').click(function(){
     toggle_form_fields(id_protocol());
-    // update_issuer();
   });
 
   // $('input[name="service_provider[issuer_department]"]').keyup(update_issuer);
--- a/app/models/service_provider.rb
+++ b/app/models/service_provider.rb
@@ -19,7 +19,7 @@ class ServiceProvider < ApplicationRecord
   validate :certs_are_pems
 
   enum block_encryption: { 'none' => 0, 'aes256-cbc' => 1 }, _suffix: 'encryption'
-  enum identity_protocol: { openid_connect: 0, saml: 1 }
+  enum identity_protocol: { openid_connect_private_key_jwt: 0, openid_connect_pkce: 2, saml: 1 }
 
   before_validation(on: %i[create update]) do
     self.attribute_bundle = attribute_bundle.reject(&:blank?) if attribute_bundle.present?
@@ -96,6 +96,10 @@ class ServiceProvider < ApplicationRecord
     @certificates = nil
 
     serial
+  end
+
+  def oidc?
+    openid_connect_pkce? || openid_connect_private_key_jwt?
   end
 
   private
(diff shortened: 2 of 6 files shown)


app/javascript/app/service-provider-form.js
```diff
@@ -10,17 +10,25 @@ $(function(){
 
   var toggle_form_fields = function(id_protocol){
     switch(id_protocol){
-      case 'openid_connect':
+      case 'openid_connect_private_key_jwt':
```
orenyk, line 13: I wanted to add/update specs for this but we currently don't have any 😞, or rather the specs that exist are all skipped. I'm questioning the utility of adding those specs since this whole site needs a major overhaul, but I could be convinced to try and write them.

app/javascript/app/service-provider-form.js
```diff
@@ -10,17 +10,25 @@ $(function(){
 
   var toggle_form_fields = function(id_protocol){
     switch(id_protocol){
```
aduth, line 12: There's a similar switch in the function below this one (`issuer_protocol_name`). Should that be updated? Or removed? I can't see that it's even being used based, per the note at the top of the file about `update_issuer` being commented out.

orenyk: Good catch! I think that just updates the example issuer in the help text but that would also be impacted. I'll update, thanks!

orenyk: I updated but then looking further it looks like this JS is actually unused. I think it makes sense to update the `switch` anyway just in case someone turns it on but we can remove it in a larger redesign/refactor, if that makes sense to you. Let me know what you think!

app/javascript/app/service-provider-form.js
```diff
@@ -10,23 +10,33 @@ $(function(){
...
+        $('.cert-fields').show();
     }
   }
 
   var issuer_protocol_name = function(id_protocol){
     switch(id_protocol){
-      case 'openid_connect':
+      case 'openid_connect_private_key_jwt':
+        return 'openidconnect';
+      case 'openid_connect_pkce:
```
aduth, line 39: syntax error: this string is missing the closing quote:

```suggestion
      case 'openid_connect_pkce':
```

app/javascript/app/service-provider-form.js
aduth: > I think it makes sense to update the `switch` anyway just in case someone turns it on but we can remove it in a larger redesign/refactor, if that makes sense to you. Let me know what you think!

Normally I'd be in favor of removing the code so we don't have to deal with situations like this where we're maintaining code that isn't being run anyways. We can always recall it from git history if we need it.

I don't feel strongly though.

app/javascript/app/service-provider-form.js
orenyk: Thanks I caught that locally and updated 😄

app/javascript/app/service-provider-form.js
orenyk: hm fair enough - I'll do a quick double-check to make sure it's not being referenced and just delete it.

orenyk: all set!

app/models/service_provider.rb
```diff
@@ -19,7 +19,7 @@ class ServiceProvider < ApplicationRecord
   validate :certs_are_pems
 
   enum block_encryption: { 'none' => 0, 'aes256-cbc' => 1 }, _suffix: 'encryption'
-  enum identity_protocol: { openid_connect: 0, saml: 1 }
```
aduth, line 22: There's a reference to `openid_connect` here:

https://github.com/18F/identity-dashboard/blob/889a03b6ee324c271dd30f120f118380bae11dc3/app/views/service_providers/show.html.erb#L92

Do we need to update it?

Do we have to worry about migrating any existing data with this change?

orenyk: All the good questions! Migrating data is not necessary since the database is just storing integer representations (although I should check if we were using any of the helper methods, e.g. `ServiceProvider#openid_connect?` - don't think so assuming our test coverage is _decent_).

That said, yes, that logic will break so I will fix it. Thanks for the close look!

orenyk: Oh, there are a few existing SPs that we'd probably want to move over (at least the ones in prod already) - I can find them and make the update once this is deployed.

app/javascript/app/service-provider-form.js
```diff
@@ -10,53 +10,36 @@ $(function(){
...
-      case 'openid_connect':
+      case 'openid_connect_private_key_jwt':
         $('.saml-fields').hide();
         $('.oidc-fields').show();
+        $('.cert-fields').show();
+        break;
+      case 'openid_connect_pkce':
+        $('.saml-fields').hide();
+        $('.oidc-fields').show();
+        $('.cert-fields').hide();
```
zachmargolis, line 21: if an SP that uses PKCE ever wants to submit RISC events to our backend, they'll need a cert...is there a way to show the certs but mark them as optional?

zachmargolis (approved):
LGTM

app/javascript/app/service-provider-form.js
orenyk: Whoa good catch, did not realize that. I'll add some text to that effect to the field label. Thanks!

app/views/service_providers/show.html.erb
```diff
@@ -90,10 +90,9 @@
 <h2><label for="push_notification_url">Push Notification URL:</label></h2>
 <p class="font-mono-xs margin-top-0" name="push_notification_url"><%= service_provider.push_notification_url %></p>
 
-    <% if service_provider.identity_protocol == 'openid_connect' %>
+    <% if service_provider.oidc? %>
```
achapm, line 93: I suggest we get rid of this condition and call this "Additional Redirect URIs" for OIDC and SAML

achapm (approved):
Suggestion above, otherwise LGTM

app/views/service_providers/show.html.erb
orenyk: I don't mind keeping this for now, I think the language is clearer and this whole page will eventually get refactored so we can improve it then. Thanks though!

PR #457, opened 2021-06-22
EventBridge API fixes

zachmargolis:
Discovered from testing in `dev`

### DIFF
--- a/app/services/risc_destination_updater.rb
+++ b/app/services/risc_destination_updater.rb
@@ -25,7 +25,7 @@ class RiscDestinationUpdater
   end
 
   def issuer_slug
-    service_provider.issuer.gsub(/[^.\-_A-Za-z0-9]/, '_')
+    @issuer_slug ||= service_provider.issuer.gsub(/[^.\-_A-Za-z0-9]/, '_')
   end
 
   def connection_name
--- a/app/services/risc_destination_updater.rb
+++ b/app/services/risc_destination_updater.rb
@@ -72,7 +72,6 @@ class RiscDestinationUpdater
       state: 'ENABLED',
       description: "Rule for for #{service_provider.friendly_name}",
       event_bus_name: event_bus_name,
-      role_arn: destination_iam_role_arn,
     )
   end
 
@@ -82,6 +81,7 @@ class RiscDestinationUpdater
       event_bus_name: event_bus_name,
       targets: [
         id: target_id,
+        role_arn: destination_iam_role_arn,
         arn: api_destination_arn,
         input_path: '$.detail',
         http_parameters: {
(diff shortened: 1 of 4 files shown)


app/services/risc_destination_updater.rb
```diff
@@ -24,27 +24,36 @@ def remove
     eventbridge_client.delete_connection(name: connection_name) if connection_exists?
   end
 
+  def issuer_slug
+    service_provider.issuer.gsub(/[^.\-_A-Za-z0-9]/, '_')
```
stevegsa, line 28: perhaps some memoization?

stevegsa (approved):
lgtm

app/services/risc_destination_updater.rb
aduth, line 28: Could use [`ActiveSupport::Inflector.parameterize`](https://api.rubyonrails.org/v6.1.3.2/classes/ActiveSupport/Inflector.html#method-i-parameterize) here?

```suggestion
    ActiveSupport::Inflector.parameterize(service_provider.issuer, separator: '_')
```

Or if we'd be fine with dashes as the separator:

```suggestion
    service_provider.issuer.parameterize
```

app/services/risc_destination_updater.rb
```diff
@@ -24,27 +24,36 @@ def remove
...
   end
 
   def event_bus_name
     # Matches value managed by Terraform
     "#{Identity::Hostdata.env}-risc-notifications"
   end
 
+  def destination_iam_role_arn
+    # Matches value managed by Terraform
+    "arn:aws:iam::#{aws_account_id}:role/#{Identity::Hostdata.env}-risc-notification-destination"
```
aduth, line 54: Any way we could incorporate a test assertion here? Partly also wondering if `::` is expected?

aduth (approved):

app/services/risc_destination_updater.rb
zachmargolis: This regex came from inverting the regex in error message I got from the API through trial and error, so I figure it's easier to maintain as-is?

app/services/risc_destination_updater.rb
zachmargolis: so it turns out... the `::` is not expected it was a copy-paste error on my part, but yes will add a test case

app/services/risc_destination_updater.rb
zachmargolis: done in c4fa91c00ef46d11e4ee9b63bd5eb16c1a9b765e

app/services/risc_destination_updater.rb
zachmargolis: I got some ARNs confused... I believe this *should* have the `::`... added a spec in 89763b0fa39693d2dafb7fd76f1a897fa2e733f8, [slack message why i think this is right](https://gsa-tts.slack.com/archives/C01D7MWNE64/p1624397530017100?thread_ts=1624374211.009400&cid=C01D7MWNE64)

PR #453, opened 2021-06-14
[LG-4666] Add email NameID format config to SPs

orenyk:
Resolves LG-4666

* Add email_nameid_format_allowed column to service_providers table
* Update strong params and add controls to edit and view the attribute
  if you're an admin
* Add basic feature specs

PR #458, opened 2021-06-23
[LG-4740] Actually send email_nameid_format_allowed to IdP

orenyk:
Resolves LG-4740

Whoops left this out of #453


achapm (approved):
LGTM

PR #460, opened 2021-06-24
Update EventBridge names to fit within 64 character limit

zachmargolis:
Ran into another issue testing in my env

I wish we had another stable identifier for these besides DB ID, but this is what we have for now

Luckily the IDP does not need to reference these IDs (because they'd be different on the IDP side anyways), they just need to be consistent from the side of whatever manages these, which is Dashboard vs Dashboard.

### DIFF
--- a/app/services/risc_destination_updater.rb
+++ b/app/services/risc_destination_updater.rb
@@ -25,7 +25,7 @@ class RiscDestinationUpdater
   end
 
   def issuer_slug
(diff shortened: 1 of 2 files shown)


mitchellhenke (approved):

orenyk (approved):
looks good, although I'm a little fuzzy on the broader context 😄

PR #461, opened 2021-06-25
Stop AWS S3 from calculating MD5 checksums, not FIPS-supported

zachmargolis:


app/controllers/service_providers_controller.rb
```diff
@@ -194,7 +194,11 @@ def using_s3?
   end
 
   def s3
-    @s3 ||= Aws::S3::Client.new(region: IdentityConfig.store.aws_region)
+    @s3 ||= Aws::S3::Client.new(
+      region: IdentityConfig.store.aws_region,
+      signature_version: 'v4',
+      compute_checksums: false,
```
mitchellhenke, line 200: is `compute_checksums: false` needed if `signature_version` is set to v4?

zachmargolis: so.... I couldn't actually find any documentation on `signature_version` (which @crissupb mentioned...) I could only see `compute_checksums`? 

https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#initialize-instance_method

🤷

zachmargolis:
Also, note, we may want to wait for https://github.com/18F/identity-hostdata/pull/31 to merge and then include this

app/controllers/service_providers_controller.rb
zachmargolis: followed up, default signature is `v4`, so I removed that and left in `compute_checksums`

zachmargolis:
admin overriding the diff coverage issue

PR #462, opened 2021-06-28
Use correct method name, add test coverage

zachmargolis:
...and that's why you always add test coverage

always leave a note


mitchellhenke (approved):

solipet (approved):
lgtm

PR #467, opened 2021-08-12
[LG-4882] Update dashboard user attribute list

TiannaW24:
https://cm-jira.usa.gov/browse/LG-4882

* Add verified_at back to the user attribute list
* Add a PR template


pull_request_template.md
```diff
@@ -0,0 +1,13 @@
...
+
+### Changes
+
+-   Add a description of your changes here
+
+### PR Checklist:
+
+1. [ ] Have you linted and tested your code locally prior to submission?
+2. [ ] Have you tagged the appropriate dev(s) for review?
+3. [ ] Have you moved any Jira/Smartsheet tickets to Code Review?
```
zachmargolis, line 13: 1. does this need to be in a `.github` directory?
2. can we maybe add it in a separate PR?

pull_request_template.md
```diff
@@ -0,0 +1,13 @@
+### Work Ticket
```
orenyk, line 1: **suggestion:** can we raise this in Engineer Huddle next week since we're not the only team that works in this repo? I assume everyone would be happy with it but I'd want to give the other devs a chance to provide feedback. I'd hold off on adding it in this PR and we can always add it separately.

app/models/service_provider.rb
```diff
@@ -64,6 +64,7 @@ def self.possible_attributes
       phone
       x509_subject
       x509_presented
+      verified_at
```
zachmargolis, line 67: can we add a spec to make sure this gets used where we think it does? like a view spec or a controller spec?

orenyk, line 67: **question:** Do we have any tests for changing redirect_uris? If we do, can we modify the tests to check this off as well (so we implicitly test the fact that this attribute is available)?

orenyk (approved):
One question re: testing and a suggestion to hold on adding the PR template, otherwise nice job!

spec/features/service_providers_spec.rb
```diff
@@ -23,6 +23,7 @@
 
       check 'email'
       check 'first_name'
+      check 'verified_at'
```
TiannaW24, line 26: @orenyk @zachmargolis updated these tests to make sure verified_at is available and selectable

zachmargolis (approved):
LGTM! thanks for adding the specs

spec/features/service_providers_spec.rb
orenyk: Great!

orenyk (approved):
:shipit:

PR #470, opened 2021-08-16
[LG-4713] Adding a team link to the service provider page

jauschalley:
Instead of just display text for the team name, we are now creating a link to the team page.  
I also made some adjustments to tests that needed to included the team during app creation to stage the test realistically.

### DIFF
--- a/app/views/service_providers/show.html.erb
+++ b/app/views/service_providers/show.html.erb
@@ -1,5 +1,4 @@
 <%= link_to 'Back to apps', '/service_providers', :class => 'usa-link' %>
-<br>
 
 <h1 class="usa-display">Details for "<%= service_provider.friendly_name %>"</h1>
 
--- a/app/views/service_providers/show.html.erb
+++ b/app/views/service_providers/show.html.erb
@@ -12,7 +12,9 @@
 <p class="font-mono-xs margin-top-0" name="agency"><%= service_provider&.agency&.name %></p>
 
 <h2><label for="team">Team:</label></h2>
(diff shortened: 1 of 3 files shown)


app/views/service_providers/show.html.erb
```diff
@@ -13,7 +12,7 @@
 <p class="font-mono-xs margin-top-0" name="agency"><%= service_provider&.agency&.name %></p>
 
 <h2><label for="team">Team:</label></h2>
-<p class="font-mono-xs margin-top-0" name="agency"><%= service_provider.team %></p>
+<%= link_to "#{service_provider&.team}", team_path(service_provider&.team&.id), :class => 'usa-link' %>
```
orenyk, line 15: **suggestion:** I _think_ you might be able to make this work by just linking to the model. What do you think about wrapping this in a conditional to avoid all the nil checks? This should also allow us to leave the tests alone.

```suggestion
<% if service_provider.team.present? %>
  <%= link_to service_provider.team, service_provider.team, :class => 'usa-link' %>
  <% end %>
```

P.S. Rails magic is magical

orenyk (approved):
One small suggestion (that I think should allow you to remove all of the changes to the tests) - otherwise looks good!

app/views/service_providers/show.html.erb
zachmargolis: Also, `"#{service_provider&.team}"` the `#{}` is not needed since the entirety of the string is the thing inside the interpolation

PR #474, opened 2021-09-02
[LG-5012] updating yaml blobs per protocol

jauschalley:
Updating the yaml blobs on the service provider pages making them easier to use for integrations.  The blob shape is conditional per the protocol.  The code is a little long-winded but I opted for self-documenting and easy to make simple changes in the future.

https://cm-jira.usa.gov/browse/LG-5012


spec/factories/service_providers.rb
```diff
@@ -15,6 +15,14 @@
       identity_protocol { 1 }
     end
 
+    trait :with_oidc_jwt do
+      identity_protocol { 0 }
+    end
+
+    trait :with_oidc_pke do
+      identity_protocol { 2 }
+    end
```
zachmargolis, line 24: 1. I am pretty sure we can assign the enum names here for readability:
2. should that be pk_c_e?

```suggestion
    trait :with_oidc_jwt do
      identity_protocol { :openid_connect_private_key_jwt }
    end

    trait :with_oidc_pcke do
      identity_protocol { :openid_connect_pkce }
    end
```

spec/helpers/service_provider_helper_spec.rb
```diff
@@ -64,29 +64,79 @@
...
     end
+  end
 
-    it 'returns the sp configuration in the yaml blurb' do
-      sp.attribute_names.each do |attribute_name|
-        next if sp_config_protected_attributes.include?(attribute_name)
-        expect(yamlized_sp(sp)).to include(attribute_name)
+  describe '#config_hash' do
+    let(:saml_sp) { create(:service_provider, :saml) }
+    let(:oidc_pke_sp) {create(:service_provider, :with_oidc_pke )}
```
zachmargolis, line 75: pke ➡️ pkce

```suggestion
    let(:oidc_pkce_sp) {create(:service_provider, :with_oidc_pkce )}
```

spec/helpers/service_provider_helper_spec.rb
```diff
@@ -64,29 +64,79 @@
...
+    end
+    it 'returns a properly formatted yaml blurb for SAML' do
+      sp_config_saml_attributes.each do |attribute_name|
+        expect(config_hash(saml_sp)).to include(attribute_name)
+      end
+    end
+    it 'returns a properly formatted yaml blurb for OIDC pke' do
+      sp_config_oidc_attributes.each do |attribute_name|
+        expect(config_hash(oidc_pke_sp)).to include(attribute_name)
+      end
```
zachmargolis, line 135: pke - PKCE

```suggestion
    it 'returns a properly formatted yaml blurb for OIDC pkce' do
      sp_config_oidc_attributes.each do |attribute_name|
        expect(config_hash(oidc_pkce_sp)).to include(attribute_name)
      end
```

orenyk (changes requested):
Nice job! I left a few comments suggesting some ways to keep the different logic branches slimmed down (that you're more than welcome to disagree with) and raised a small issue around the protocol, but it's definitely close. Let me know if you have any questions!

app/helpers/service_provider_helper.rb
```diff
@@ -128,4 +131,56 @@ def formatted_config_hash(sp_json)
     end.to_h
   end
   # rubocop:enable Layout/LineLength
+
+  def map_oidc_display_configs(sp_hash, sp_id, agency = {name => '', id => nil})
```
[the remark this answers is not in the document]
jauschalley: I ended up conceding on this point.  By the time all the paths were built in, it started to make much more sense to me to have the base hash and then manipulate that hash per each conditional.

spec/factories/service_providers.rb
```diff
@@ -12,7 +12,19 @@
...
+    trait :with_oidc_jwt do
+      identity_protocol { :openid_connect_private_key_jwt }
+    end
+
+    trait :with_oidc_pkce do
+      identity_protocol { :openid_connect_pkce }
+    end
+
+    trait :with_ial_2 do
+      ial {2}
```
orenyk, line 27: **nitpick:** I think we try to keep spaces around literal hashes:
```suggestion
      ial { 2 }
```

app/helpers/service_provider_helper.rb
```diff
@@ -128,4 +131,56 @@ def formatted_config_hash(sp_json)
...
+      'logo' => '<REPLACE_ME.png>',
+      'certs' => '<REPLACE_ME>',
+      'return_to_sp_url' => sp_hash['return_to_sp_url'],
+      'redirect_uris' => sp_hash['redirect_uris'],
+      'return_to_sp_url' => sp_hash['return_to_sp_url'],
+      'failure_to_proof_url' => sp_hash['failure_to_proof_url'],
+      'ial' => sp_hash['ial'],
+      'attribute_bundle' => sp_hash['attribute_bundle'],
+      'restrict_to_deploy_env' => 'prod',
+      'protocol' => sp_hash['protocol'],
```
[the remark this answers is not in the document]
orenyk: **issue:** follow up - the raw `identity_protocol` attribute won't be quite right since there are three options permitted (`saml`, `openid_connect_private_key_jwt`, and `openid_connect_pkce`), but we want it to either be `oidc` or `saml` exactly. What do you think about setting this attribute explicitly in each of the protocol-specific methods (e.g. overriding what comes in as part of the `config_hash`)?

app/helpers/service_provider_helper.rb
```diff
@@ -128,4 +127,54 @@ def formatted_config_hash(sp_json)
...
+    }
+    hash_with_ial_attr = add_IAL_attribute(base_hash, sp_hash['failure_to_proof_url'])
+    if hash_with_ial_attr['protocol'] == 'saml'
+      add_saml_attributes(hash_with_ial_attr)
+    else
+      add_pkce_atttribute(hash_with_ial_attr)
+    end
+  end
+
+  # rubocop:disable Layout/LineLength
```
orenyk, line 159: **suggestion:** I generally like overriding rubocop specifically on the lines that violate the linter, so I'd recommend wrapping line 163 instead of the whole method.

orenyk (approved):
Looks good! One more change needed around the `protocol` but good to merge once that's resolved. Nice work!

app/helpers/service_provider_helper.rb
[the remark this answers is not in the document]
jauschalley: I definitely didn't read that first comment correctly.

PR #476, opened 2021-09-09
Use login.gov shared ESLint configuration

aduth:
See: https://github.com/18F/identity-idp/pull/5359

**Why:** So that we can use a common standard ruleset across all repositories where JavaScript is used.

Notes:

- Doesn't yet incorporate Prettier, since that will be a significant number of changes that can be safely be applied separately.
- Removes PostCSS configuration. This follows a [similar removal in IdP](https://github.com/18F/identity-idp/pull/4260), and we don't appear to be using [import inlining](https://github.com/postcss/postcss-import). Flexbugs and env settings seem most applicable to legacy browsers like Internet Explorer, but we appear to already be targeting newer browsers with e.g. [unpolyfilled usage of modern JavaScript like String#includes](https://github.com/18F/identity-dashboard/blob/abb55d8efa906f3db62ffa93f02b2d52e1bfacad/app/javascript/app/service-provider-form.js#L76).A quick check in Internet Explorer does not reveal any obvious breakage where we depend on these transforms.


zachmargolis (approved):
LGTM

aduth:
@russbeye I was going to wait for #475 to be merged to deal with the inevitable merge conflicts here. Do you know if that'll be merged soon, or would you be okay if I merge this now and you rebase and solve the conflicts yourself in #475? Happy to help with that, of course.

russbeye:
@aduth Feel free to merge away, as we've been torn with other support work for the time being. I can handle resolving any issues past that no problem. Thanks for the heads up!

aduth:
Great, thanks @russbeye 🙏

PR #477, opened 2021-09-10
Correcting yaml blob conditional attributes and ial number as an integer

jauschalley:
This is correcting some mistakes not covered in #474 . 

- IAL should now be set to a integer
- Conditional attributes for SAML v OIDC should now be populated with the correct data

### DIFF
--- a/spec/helpers/service_provider_helper_spec.rb
+++ b/spec/helpers/service_provider_helper_spec.rb
@@ -146,6 +146,9 @@ describe ServiceProviderHelper do
         expect(config_hash(oidc_jwt_sp_2)).to include(attribute_name)
       end
     end
+    it 'returns the ial config as an integer instead of a string' do 
+      expect(config_hash(saml_sp_ial_2)['ial'].class).to eq(Integer)
+    end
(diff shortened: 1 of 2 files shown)


TiannaW24 (approved):

spec/helpers/service_provider_helper_spec.rb
```diff
@@ -146,6 +146,9 @@
         expect(config_hash(oidc_jwt_sp_2)).to include(attribute_name)
       end
     end
+    it 'returns the ial config as an integer instead of a string' do 
+      expect(config_hash(saml_sp_ial_2)['ial'].class).to eq(Integer)
```
orenyk, line 150: **nitpick:** I think you can use the assertion `be_an` to clean this up a little:
```suggestion
      expect(config_hash(saml_sp_ial_2)['ial']).to be_an(Integer)
```

orenyk (approved):
:shipit:

spec/helpers/service_provider_helper_spec.rb
jauschalley: Yes.  I agree.

PR #478, opened 2021-09-15
[LG-5016] Add app name as a required field to Service Provider

TiannaW24:
Work Ticket: https://cm-jira.usa.gov/browse/LG-5016

Changes:
* Add app_name as attr to Service Provider model
* Add app_name to Dashboard views
* Create DB migration
* Update unit tests accordingly

### DIFF
--- a/app/controllers/service_providers_controller.rb
+++ b/app/controllers/service_providers_controller.rb
@@ -137,6 +137,7 @@ class ServiceProvidersController < AuthenticatedController
       :signed_response_message_requested,
       :sp_initiated_login_url,
       :logo_file,
+      :app_name,
       attribute_bundle: [],
       redirect_uris: [],
       help_text: {},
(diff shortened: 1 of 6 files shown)


russbeye (approved):

app/controllers/service_providers_controller.rb
```diff
@@ -137,6 +137,7 @@ def service_provider_params
       :signed_response_message_requested,
       :sp_initiated_login_url,
       :logo_file,
+      :app_name,
       attribute_bundle: [],
```
jauschalley, line 141: This is the name as it appears in the IAA

app/models/service_provider.rb
```diff
@@ -105,6 +107,10 @@ def oidc?
 
   private
 
+  def set_app_name_default
+    self.app_name = self.friendly_name
+  end
+
```
jauschalley, line 113: Issue: Leave blank if user does not supply name

jauschalley (approved):

PR #479, opened 2021-09-20
Remove app_id assignment in yaml blob

jauschalley:
Simply removing the app_id assignment and having the `restrict_to_deploy_env` variable display with quotes in the yaml blob.  There will be more changes with a serializer coming up but due to the glacier development pace on the partnerships team I am not sure when I'll get to it.  I think that this will cause errors if I don't fix it now and the rest is more cosmetic rather than functional.


russbeye (approved):
This helps!

PR #498, opened 2021-11-13
 [LG-5364] Adjustments to attribute validations

jauschalley:
### Work Ticket
https://cm-jira.usa.gov/browse/LG-5364

### Changes

-  We only want to validate empty attribute bundles for SAML with IAL2
-  Hiding return_to_app_url on the show page for OIDC service providers
- Added more specific tests for validations
- Added more specific error messages

### PR Checklist:

### DIFF
--- a/app/models/service_provider.rb
+++ b/app/models/service_provider.rb
@@ -159,19 +159,17 @@ class ServiceProvider < ApplicationRecord
 
   def validate_attribute_bundle
     # attribute bundle should not be empty when saml and ial2 are selected
-    if ((attribute_bundle || []).empty? && ial == 2 && identity_protocol == 'saml')
+    if !attribute_bundle.present? && ial == 2 && identity_protocol == 'saml'
       errors.add(:attribute_bundle, 'Attribute bundle cannot be empty')
       return false
     end
 
-    return true if !attribute_bundle
-
-    if contains_invalid_attribute?
+    if attribute_bundle.present?  && contains_invalid_attribute?
       errors.add(:attribute_bundle, 'Contains invalid attributes')
       return false
     end
 
-    if ial == 1 && ((attribute_bundle & ALLOWED_IAL2_ATTRIBUTES).length > 0)
+    if ial == 1 && ((attribute_bundle & ALLOWED_IAL2_ATTRIBUTES).present?)
       errors.add(:attribute_bundle, 'Contains ial 2 attributes when ial 1 is selected')
     end
     true
@@ -179,9 +177,6 @@ class ServiceProvider < ApplicationRecord
 
   def contains_invalid_attribute? 
     possible_attributes = ALLOWED_IAL1_ATTRIBUTES + ALLOWED_IAL2_ATTRIBUTES
-    attribute_bundle.each{ |att| 
-      return true if !possible_attributes.include? att
-    }
-    false
+    attribute_bundle.any? { |att| !possible_attributes.include?(att) }
   end
 end
--- a/app/models/service_provider.rb
+++ b/app/models/service_provider.rb
@@ -158,15 +158,30 @@ class ServiceProvider < ApplicationRecord
   end
 
   def validate_attribute_bundle
-    if (attribute_bundle || []).empty?
+    # attribute bundle should not be empty when saml and ial2 are selected
+    if ((attribute_bundle || []).empty? && ial == 2 && identity_protocol == 'saml')
       errors.add(:attribute_bundle, 'Attribute bundle cannot be empty')
       return false
     end
 
-    possible_attributes = ALLOWED_IAL1_ATTRIBUTES + ALLOWED_IAL2_ATTRIBUTES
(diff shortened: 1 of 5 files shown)


app/models/service_provider.rb
```diff
@@ -158,15 +158,30 @@ def certs_are_pems
...
-    return if (attribute_bundle - ALLOWED_IAL1_ATTRIBUTES).empty?
+    return true if !attribute_bundle
+
+    if contains_invalid_attribute?
+      errors.add(:attribute_bundle, 'Contains invalid attributes')
+      return false
+    end
 
-    errors.add(:attribute_bundle, 'Contains invalid IAL attributes')
+    if ial == 1 && ((attribute_bundle & ALLOWED_IAL2_ATTRIBUTES).length > 0)
```
zachmargolis, line 174: I would change this to:
```suggestion
    if ial == 1 && (attribute_bundle & ALLOWED_IAL2_ATTRIBUTES).present?
```
because it looks like we had `attribute_bundle || []` above which implies `attribute_bundle` might be nil, and:

```
nil & []
=> false
```
and `false` won't have a `#length` method

app/models/service_provider.rb
```diff
@@ -158,15 +158,30 @@ def certs_are_pems
   end
 
   def validate_attribute_bundle
-    if (attribute_bundle || []).empty?
+    # attribute bundle should not be empty when saml and ial2 are selected
+    if ((attribute_bundle || []).empty? && ial == 2 && identity_protocol == 'saml')
```
zachmargolis, line 162: style nit, usually we don't add the extra parens if we don't need them
```suggestion
    if (attribute_bundle || []).empty? && ial == 2 && identity_protocol == 'saml'
```

app/models/service_provider.rb
```diff
@@ -158,15 +158,30 @@ def certs_are_pems
...
+    end
+    true
+  end
+
+  def contains_invalid_attribute? 
+    possible_attributes = ALLOWED_IAL1_ATTRIBUTES + ALLOWED_IAL2_ATTRIBUTES
+    attribute_bundle.each{ |att| 
+      return true if !possible_attributes.include? att
+    }
+    false
```
zachmargolis, line 185: a few ideas here:
1. the spacing/formatting is off (usually we use `do...end` for multiline blocks)
2. I think we can clean this up a tad with `#all` or `#any?` which both have similar early-abort logic internally

```suggestion
    attribute_bundle.any? { |att| !possible_attributes.include?(att) }
```

zachmargolis (approved):
LGTM, some small comments

TiannaW24 (approved):

app/models/service_provider.rb
jauschalley: I had a nil check for attribute_bundle on line 167 but this is still good advice and allow me to cleanup that line, too.

jauschalley:
@zachmargolis Thanks for the suggestions.  I've cleaned it up further if you wouldn't mind taking another look.

app/models/service_provider.rb
```diff
@@ -158,15 +158,25 @@ def certs_are_pems
...
+    # attribute bundle should not be empty when saml and ial2 are selected
+    if !attribute_bundle.present? && ial == 2 && identity_protocol == 'saml'
       errors.add(:attribute_bundle, 'Attribute bundle cannot be empty')
       return false
     end
 
-    possible_attributes = ALLOWED_IAL1_ATTRIBUTES + ALLOWED_IAL2_ATTRIBUTES
-    return if ial == 2 && (attribute_bundle - possible_attributes).empty?
-    return if (attribute_bundle - ALLOWED_IAL1_ATTRIBUTES).empty?
+    if attribute_bundle.present?  && contains_invalid_attribute?
```
zachmargolis, line 167:
```suggestion
    if attribute_bundle.present? && contains_invalid_attribute?
```
extra space snuck in there

app/models/service_provider.rb
```diff
@@ -158,15 +158,25 @@ def certs_are_pems
...
+      errors.add(:attribute_bundle, 'Contains invalid attributes')
+      return false
+    end
 
-    errors.add(:attribute_bundle, 'Contains invalid IAL attributes')
+    if ial == 1 && ((attribute_bundle & ALLOWED_IAL2_ATTRIBUTES).present?)
```
zachmargolis, line 172: extra parens
```suggestion
    if ial == 1 && (attribute_bundle & ALLOWED_IAL2_ATTRIBUTES).present?
```

zachmargolis (approved):
LGTM! some small nits otherwise good to go

app/models/service_provider.rb
jauschalley: Ah caught it before I did!

jauschalley: I've become very dependent on linters...

app/models/service_provider.rb
jauschalley: lol I'm sorry.  It's embarrassing at this point.

PR #499, opened 2021-11-15
adding default_AAL to yaml blob

jauschalley:
### Work Ticket
I did not make a bug.

### Changes

-   I noticed that we do not have default_AAL in the yaml blob
- I will follow up with tests but I did want this to get kicked down the road with work interruption

### PR Checklist:


orenyk (approved):
:shipit:

PR #501, opened 2021-11-17
Add missing styles to saml fields on service provider show page

jauschalley:
### Work Ticket
https://cm-jira.usa.gov/browse/LG-5017

### Changes

-   Add missing styles to saml fields on service provider show page

### PR Checklist:

### DIFF
--- a/app/views/service_providers/show.html.erb
+++ b/app/views/service_providers/show.html.erb
@@ -108,29 +108,44 @@
 <hr>
 
 <% if service_provider.identity_protocol == 'saml' %>
-  <h2 class="margin-bottom-0"><label for="acs_url">Assertion Consumer Service URL:</label> <%= service_provider.acs_url %></h2>
+  <h2 class="margin-bottom-0">
+    <label for="acs_url">Assertion Consumer Service URL:</label> 
+    <span class="font-mono-xs margin-left-5"><%= service_provider.acs_url %></span>
+  </h2>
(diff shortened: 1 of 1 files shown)


app/views/service_providers/show.html.erb
```diff
@@ -108,29 +108,44 @@
 <hr>
 
 <% if service_provider.identity_protocol == 'saml' %>
-  <h2 class="margin-bottom-0"><label for="acs_url">Assertion Consumer Service URL:</label> <%= service_provider.acs_url %></h2>
+  <h2 class="margin-bottom-0">
```
orenyk, line 111: **question:** you sure this isn't worth turning into a partial? There's a lot of repetition. I also don't know if you need the `for` here since it isn't part of a form so you'd really just need to pass in the label text and the value.

orenyk (approved):
one thought (feel free to disregard) but otherwise :shipit:!

app/views/service_providers/show.html.erb
jauschalley: I kept most of the html alone rather and just added some styles.  It might look like a huge diff but it's not.  I typically only like to use partials if they are going to be reused though I know that isn't a universal opinion.  I feel like it just makes it more difficult to look at .erbs when the code is spread out.

jauschalley: As for the `for`, I think they could probably be removed everywhere.

PR #505, opened 2022-01-10
[LG-5365] Add Production Configuration indicator to SP Dashboard

TiannaW24:
### Work Ticket
https://cm-jira.usa.gov/browse/LG-5365

### Changes

-   Add a boolean value for indicating whether a configuration is meant for production
- Persist boolean value to ServiceProviders table

### PR Checklist:

### DIFF
--- a/app/views/service_providers/_form.html.erb
+++ b/app/views/service_providers/_form.html.erb
@@ -2,7 +2,7 @@
 <fieldset class='usa-fieldset'>
 
   <legend class='usa-sr-only'>Production Configuration</legend>
-  <%= form.label :admin,
+  <%= form.label :prod_config,
                  label: 'Production Configuration',
                  class: 'usa-label',
                  hint: I18n.t('service_provider_form.prod_config').html_safe %>
--- a/app/views/service_providers/_form.html.erb
+++ b/app/views/service_providers/_form.html.erb
@@ -1,5 +1,21 @@
 <%= form.error_notification %>
 <fieldset class='usa-fieldset'>
+
+  <legend class='usa-sr-only'>Production Configuration</legend>
+  <%= form.label :admin,
+                 label: 'Production Configuration',
+                 class: 'usa-label',
+                 hint: I18n.t('service_provider_form.prod_config').html_safe %>
+  <%= form.collection_radio_buttons(:active,
+                                    { "Yes" => true, "No" => false },
+                                    :last, :first,
+                                    collection_wrapper_tag: :ul,
+                                    collection_wrapper_class: 'usa-input-list',
+                                    item_wrapper_tag: :li) do |b| %>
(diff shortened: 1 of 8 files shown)


TiannaW24:
Screen Shot 2022-01-10 at 2 33 38 PM

TiannaW24:
Screen Shot 2022-01-10 at 2 35 31 PM

jauschalley (approved):
You may want to add a test for the new field in `spec/features/service_providers_spec.rb` but I won't block you over it. 🤷‍♂️

app/views/service_providers/_form.html.erb
```diff
@@ -1,5 +1,22 @@
 <%= form.error_notification %>
 <fieldset class='usa-fieldset'>
+
+  <legend class='usa-sr-only'>Production Configuration</legend>
+  <%= form.label :admin,
```
orenyk, line 5: **question:** should this be `form.label :prod_config`? My Rails form helpers / SimpleForm is a little fuzzy, so I could be off, but it seems like it should be the attribute name?

```suggestion
  <%= form.label :prod_config,
```

db/migrate/20220110190210_add_prod_configto_sp_table.rb
```diff
@@ -0,0 +1,5 @@
+class AddProdConfigtoSpTable < ActiveRecord::Migration[6.1]
+  def change
+    add_column :service_providers, :prod_config, :boolean
```
orenyk, line 3: **suggestion:** do we want to set a default (presumably `false`) and also prevent null values?

app/views/service_providers/show.html.erb
```diff
@@ -138,7 +145,7 @@
 
   <h2 class="margin-bottom-0">
     <label for="signed_response_message_requested">Signed Response Message Requested:</label> 
-    <span class="font-mono-xs margin-left-5"><%= service_provider.signed_response_message_requested? %></span>
+    <span class="font-mono-xs margin-left-5"><%= service_provider.signed_response_message_requested? ? 'Yes' : 'No' %></span>
```
orenyk, line 148: **praise:** big fan of this!

orenyk (approved):
Two small comments, otherwise LGTM!

db/migrate/20220110190210_add_prod_configto_sp_table.rb
jauschalley: +1 on this comment

TiannaW24: Makes sense to me.

TiannaW24:
> You may want to add a test for the new field in `spec/features/service_providers_spec.rb` but I won't block you over it. 🤷‍♂️

I'll leave the unit testing for when we have logic around the button

PR #506, opened 2022-01-14
[LG-5327] clear input formatting

jauschalley:
### Work Ticket
https://cm-jira.usa.gov/browse/LG-5327

### Changes

-   Adding controller method to clear leading and trailing whitespaces to user supplied data in the service provider form

### PR Checklist:

### DIFF
--- a/app/controllers/service_providers_controller.rb
+++ b/app/controllers/service_providers_controller.rb
@@ -216,9 +216,7 @@ class ServiceProvidersController < AuthenticatedController
     ]
 
     service_provider.attributes.each do |k,v|
-      if string_attributes.include?(k)
-        v.try(:strip!)
-      end
+      v.try(:strip!) unless !string_attributes.include?(k)
     end
     service_provider
   end
--- a/app/controllers/service_providers_controller.rb
+++ b/app/controllers/service_providers_controller.rb
@@ -71,12 +71,14 @@ class ServiceProvidersController < AuthenticatedController
   end
 
   def validate_and_save_service_provider(initial_action)
-    return save_service_provider if service_provider.valid?
+    formatted_sp = clear_formatting(@service_provider)
+    return save_service_provider(formatted_sp) if formatted_sp.valid?
+
     flash[:error] = I18n.t('notices.service_providers_refresh_failed')
     render initial_action
   end
 
-  def save_service_provider
+  def save_service_provider(service_provider)
     service_provider.save!
     flash[:success] = I18n.t('notices.service_provider_saved', issuer: service_provider.issuer)
     publish_service_providers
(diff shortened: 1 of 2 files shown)


spec/controllers/service_providers_controller_spec.rb
```diff
@@ -16,6 +16,39 @@
         content_type: 'image/svg+xml',
       }
     end
+    # let(:sp_with_whitespaces) {create(:service_provider, :with_whitespaces, user: user)}
+
+    context 'when a user enters data into text inputs with leading and trailing spaces' do
```
TiannaW24, line 21: yay for unit tests!

spec/controllers/service_providers_controller_spec.rb
```diff
@@ -16,6 +16,39 @@
         content_type: 'image/svg+xml',
       }
     end
+    # let(:sp_with_whitespaces) {create(:service_provider, :with_whitespaces, user: user)}
```
TiannaW24, line 19: can delete?

russbeye, line 19: **nitpick (style):** Can we remove commented code?

russbeye (approved):

spec/controllers/service_providers_controller_spec.rb
zachmargolis, line 19: let's remove this commented out line?
```suggestion
```

app/controllers/service_providers_controller.rb
```diff
@@ -197,5 +199,27 @@ def add_iaa_warning
...
+      production_issuer
+      failure_to_proof_url
+      push_notification_url
+      app_name
+    ]
+
+    service_provider.attributes.each do |k,v|
+      v.try(:strip!) unless !string_attributes.include?(k)
+    end
+    service_provider
```
zachmargolis, line 221: a few ideas:
1. let's only iterate the string attributes?
2. I think we could just inline `string_attributes`
3. I have personal beef with `unless` so I prefer `if`
4. I'm worried that iterating the attributes hash like this won't write back to the model?

```suggestion
   %w[
      issuer
      friendly_name
      description
      metadata_url
      acs_url
      assertion_consumer_logout_service_url
      sp_initiated_login_url
      return_to_sp_url
      production_issuer
      failure_to_proof_url
      push_notification_url
      app_name
    ].each do |string_attribute|
      service_provider[string_attribute] = service_provider[string_attribute]&.strip
    end
    service_provider
```

zachmargolis (approved):
LGTM

orenyk (approved):
I like @zachmargolis's suggestions, otherwise looks great! :shipit:

app/controllers/service_providers_controller.rb
jauschalley: I also have a personal beef with unless but rubocop forced me into this.

zachmargolis: !!! let's remove that rule, we removed it in the IDP a while ago

jauschalley: I don't think we can iterate over `string_attributes` because all of them will not exist on each service provider depending on the protocol.  And the test checks that this method is run before an update of the record is returned with properly stripped values.  Further, just using `strip` is will not save to the model and dropping the `try` will blow up on nils which certainly exist in this array of keys depending on the protocol.

spec/controllers/service_providers_controller_spec.rb
jauschalley: This feels passive aggressive coming from you. 😄

app/controllers/service_providers_controller.rb
zachmargolis: AFAIK, the attributes are on each SP depending on the protocol but somtimes they're nil

And the `&.` is the new ruby shorthand for `.try` so I think it handles?

PR #508, opened 2022-01-18
Add PKI service to deploy dashboard

zachmargolis:
I'm appdev oncall (aka deployer) this week, and I wanted a quick way to see if there were pending changes to the PKI app I should deploy, so I figured this was worth adding.

**Note**: Will likely need a proxy change in devops to allow these requests to work

Screen Shot 2022-01-18 at 9 53 00 AM

### DIFF
--- a/config/status_checks.yml
+++ b/config/status_checks.yml
@@ -10,6 +10,9 @@ deploys:
   - app: 'identity-saml-sinatra'
     env: 'prod'
     host: 'https://prod-identity-saml-sinatra.app.cloud.gov/'
+  - app: 'identity-pki'
+    env: 'prod'
+    host: 'https://checking-deploy.pivcac.prod.login.gov'
 
   # Staging
   - app: 'identity-idp'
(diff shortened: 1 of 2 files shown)


config/status_checks.yml
```diff
@@ -10,6 +10,9 @@ deploys:
   - app: 'identity-saml-sinatra'
     env: 'prod'
     host: 'https://prod-identity-saml-sinatra.app.cloud.gov/'
+  - app: 'identity-pki'
+    env: 'prod'
+    host: 'https://checking-deploy.pivcac.prod.login.gov'
```
zachmargolis, line 15: So we need to have *something* to go in the wildcard domain here, and while I do love `omg-r-u-srsly-expired.pivcac.` that [we use for our cert healthchecks](https://github.com/18F/identity-devops/blob/9c6068c333aa08043ffe708cdf48e9fa667bd14b/terraform/modules/newrelic/pivcac.tf#L7) I figured this was hopefully clear enough

mitchellhenke (approved):

PR #526, opened 2022-03-24
Allow more complex HTML elements in SP help text

orenyk:
### Changes

-  Moar HTML options!

### PR Checklist:

### DIFF
--- a/app/models/service_provider.rb
+++ b/app/models/service_provider.rb
@@ -118,7 +118,7 @@ class ServiceProvider < ApplicationRecord
 
   def sanitize_section(section)
     section.transform_values! do |translation|
-      sanitize translation, tags: %w[a b br p], attributes: %w[href target]
+      sanitize translation, tags: %w[a b strong em br p ol ul li], attributes: %w[href target]
     end
   end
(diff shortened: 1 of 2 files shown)


app/models/service_provider.rb
```diff
@@ -118,7 +118,7 @@ def sanitize_help_text_content
 
   def sanitize_section(section)
     section.transform_values! do |translation|
-      sanitize translation, tags: %w[a b br p], attributes: %w[href target]
+      sanitize translation, tags: %w[a b i br p ol ul li], attributes: %w[href target]
```
aduth, line 121: Could we add `strong` and `em` in addition to `b` and `i`, as the usually-more-semantically-correct option for bold and italic? Maybe leave out `i` altogether?

aduth (approved):

app/models/service_provider.rb
orenyk: sure - I had that thought but then just figured I'd follow the pattern for what was there. We def have people using `b` already so I won't remove that but happy to skip `i`.

PR #529, opened 2022-04-11
Specify ruby version

NatashaJIbrahim:
### Changes

-  Specifies the ruby version to avoid making people symlink the versions they have downloaded for their version manager to work. Since the whole purpose of a ruby version manager is to let us have multiple versions on our machine at once. 
- This should reduce friction for people setting up their machines to work in this codebase.


zachmargolis (approved):
LGTM

Previously, especially for the IDP, we encouraged aliasing ruby versions so we didn't have to worry about specific patches so much, so we only needed to worry about minor releases? Unsure if we still care

aduth:
> Previously, especially for the IDP, we encouraged aliasing ruby versions so we didn't have to worry about specific patches so much, so we only needed to worry about minor releases? Unsure if we still care

Previously: https://github.com/18F/identity-idp/pull/5819/files#r785002111

Most curious to me there is @mitchellhenke 's comment about deployed environments. Would we expect any issues from pinning to a specific patch version?

mitchellhenke:
> > Previously, especially for the IDP, we encouraged aliasing ruby versions so we didn't have to worry about specific patches so much, so we only needed to worry about minor releases? Unsure if we still care
> 
> Previously: https://github.com/18F/identity-idp/pull/5819/files#r785002111
> 
> Most curious to me there is @mitchellhenke 's comment about deployed environments. Would we expect any issues from pinning to a specific patch version?

There's no issue as long as the deployed environment has the version specified (and 2.7.5 is one of them), though moving to Ruby 3.0 would be great 🙂

PR #548, opened 2022-05-27
Start converting jQuery code to plain old fashioned javascript

jmhooper:
This commit starts converting some of the functionality in the service provider form to vanilla JS. This is intended to allow us to drop jQuery as a dep from the dashboard.

### DIFF
--- a/app/javascript/app/service-provider-form.js
+++ b/app/javascript/app/service-provider-form.js
@@ -57,7 +57,6 @@ function protocolOptionSetup() {
 
   // Functions
   const toggleSAMLOptions = () => {
-
     samlFields.forEach(showElement);
     oidcFields.forEach(hideElement);
   };
--- a/app/javascript/app/service-provider-form.js
+++ b/app/javascript/app/service-provider-form.js
@@ -7,17 +7,17 @@ function ialOptionSetup() {
 
   // Functions
   const toggleIAL1Options = () => {
-    ialAttributesCheckboxes.forEach(checkboxWrapper => {
+    ialAttributesCheckboxes.forEach((checkboxWrapper) => {
       const checkboxInput = checkboxWrapper.querySelector('input');
       if (ial1Attributes.includes(checkboxInput.value)) {
         checkboxWrapper.classList.add('display-none');
-        checkboxInput.checked = false
+        checkboxInput.checked = false;
       }
     });
   };
 
   const toggleIAL2Options = () => {
-    ialAttributesCheckboxes.forEach(checkboxWrapper => {
+    ialAttributesCheckboxes.forEach((checkboxWrapper) => {
       checkboxWrapper.classList.remove('display-none');
     });
   };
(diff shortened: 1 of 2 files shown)


app/javascript/app/service-provider-form.js
```diff
@@ -1,52 +1,68 @@
...
-  const ialAttributesCheckboxes = $('.ial-attr-wrapper');
-  const fileContainer = $('#certificate-container');
-  const logoInput = $('.input-file');
-  const pemInputMessage = $('.js-pem-input-error-message');
-  const pemInput = $('.js-pem-input');
-  const redirectURI = $('#add-redirect-uri-field');
-  const failureToProofURL = $('.service_provider_failure_to_proof_url');
-  const returnToSpUrl = document.querySelector('#service_provider_return_to_sp_url');
-
+  const ialLevel = document.querySelector('#service_provider_ial');
```
aduth, line 3: If querying by ID, `getElementById` is a bit more targeted, equally concise, and marginally more performant.

```suggestion
  const ialLevel = document.getElementById('service_provider_ial');
```

app/javascript/app/service-provider-form.js
```diff
@@ -59,35 +75,45 @@ $(function () {
...
+  certificateUploadSetup();
+}
+
+window.addEventListener('DOMContentLoaded', serviceProviderForm);
 
-  ialLevel.change((evt) => toggleIALOptions(evt.target.value));
+$(function () {
+  if (!$('.service-provider-form').length) {
+    return;
+  }
```
zachmargolis, line 103: the PR title is "start converting" so this is sticking around for future work?

jmhooper: Yeah, I'm gonna be out next week so I wanted to get the progress I've made up

app/javascript/app/service-provider-form.js
```diff
@@ -1,52 +1,68 @@
...
+  const activeIdProtocol = document.querySelector('input[name="service_provider[identity_protocol]"]:checked');
+  const samlFields = document.querySelectorAll('.saml-fields');
+  const oidcFields = document.querySelectorAll('.oidc-fields');
+  const returnToSpUrl = document.querySelector('#service_provider_return_to_sp_url');
+
+  // Functions
   const toggleSAMLOptions = () => {
-    samlFields.show();
-    oidcFields.hide();
+    samlFields.forEach(field => field.classList.remove('display-none'));
```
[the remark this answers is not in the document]
jmhooper: I went with something a little different, but with the same idea behind it. PTAL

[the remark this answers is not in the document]
aduth: Looks good 👍  Might think explicit hide/show could work better for singular elements, since you could avoid the extra arrow function in the `forEach` callback, e.g.

```js
const showElement = (element) => element.classList.remove('display-none');

// ...

fields.forEach(showElement);
```

jmhooper:
@aduth @zachmargolis: I did a little more cleanup on this. Either of you mind taking a look? I still need an approval.

zachmargolis (approved):
LGTM

app/javascript/app/service-provider-form.js
```diff
@@ -1,52 +1,71 @@
...
+        toggleIAL2Options();
+        break;
+      default:
+        showElement(failureToProofURL);
+        toggleIAL2Options();
+    }
   };
 
+  // Page initialization
+  toggleIALOptions(ialLevel.value);
```
aduth, line 45: Apparently we load this script on the "Show" page for an SP, triggering an error at this line. I assume it should only be loaded on the edit form itself?

```
Uncaught TypeError: Cannot read properties of null (reading 'value')
```

app/javascript/app/service-provider-form.js
```diff
@@ -1,52 +1,71 @@
...
-    ialAttributesCheckboxes.each((idx, attr) => {
-      const element = $(attr).find('input');
-
-      if (!ial1Attributes.includes(element.val())) {
-        $(attr).hide();
-        element.prop('checked', false);
+    ialAttributesCheckboxes.forEach((checkboxWrapper) => {
+      const checkboxInput = checkboxWrapper.querySelector('input');
+      if (ial1Attributes.includes(checkboxInput.value)) {
+        hideElement(checkboxWrapper);
```
aduth, line 16: I think something is backward here. In testing, "Authentication only" hides everything _except_ the IdV items.

jmhooper: Good catch, missed a `!` in the conditional. I'll fix and add a test

PR #555, opened 2022-06-08
Remove active session check and timeout logic

jmhooper:
I was trying to convert the JS for this timeout modal from jQuery to vanilla JS. I could not figure out where any of it was used so I've just removed it in this commit. After a bit of digging around in the history I'm pretty confident this code has never actually been run.


jmhooper:
Just pushed a commit to get rid of all of `utils.js`. There was some more jQuery in there for adding a warning for missing required fields on super old browsers, something that should be covered by our validations on the backend.

aduth (approved):
Nice bit of clean-up!

PR #556, opened 2022-06-08
Remove jQuery from dep list

jmhooper:
This commit gets rid of JQuery since previous work replaced the code that depended on it with plain old-fashioned JS.


aduth:
Can we remove this as well?

https://github.com/18F/identity-dashboard/blob/98b82b3027794fe6d38d8ced685ff06ddcea365d/.eslintrc.json#L9

aduth:
Are we able to remove the `jquery-rails` gem?

https://github.com/18F/identity-dashboard/blob/98b82b3027794fe6d38d8ced685ff06ddcea365d/Gemfile#L17

aduth (approved):
Excellent! Bye jQuery 👋

PR #557, opened 2022-06-08
Get rid of application.js

jmhooper:
The only thing that app.js did was included the service provider JS on every page and include the utils.js which is now gone. This commit only includes the service provider js where necessary and removes application.js.


zachmargolis (approved):
huge! LGTM

app/javascript/packs/service-provider-form.js
```diff
@@ -132,6 +132,10 @@ function logoUploadSetup() {
   const logoInput = document.querySelector('.logo-input-file');
   const logoPreview = document.querySelector('.input-preview');
 
+  if (!logoInput) {
```
aduth, line 135: Can we drop the condition on line 168 of this file, which was added in a previous pull request to support this script being loaded on pages other than the form (no longer necessary)?

app/views/layouts/application.html.erb
```diff
@@ -174,8 +174,5 @@
         </div>
       </div>
     </footer>
-
-    <%= javascript_include_tag 'js/main.js' %>
```
aduth, line 178: Did this previously not exist? 🤔

aduth (approved):
👍

app/views/layouts/application.html.erb
jmhooper: yeah, this was asking for something that does not exist

jmhooper: actually, double checking, this may be USWDS JS. Let me make sure I haven't broken everything

aduth: > `identity-style-guide` is in the asset path and this is pulling from there.

Ah, tricky!

PR #485, opened 2021-10-13
Return to app url is a conditionally required field

TiannaW24:
### Work Ticket
https://cm-jira.usa.gov/browse/LG-5083

### Changes

- Add front end changes to make return_to_sp_url a required field for all SAML integrations
- Hide return_to_sp_url for OIDC integrations and make the field not required so that SPs can submit the form
- Update unit tests
- Update field content

### PR Checklist:

1. Have you tagged the appropriate dev(s) for review?
2. Have you moved any Jira/Smartsheet tickets to Code Review?

PR #558, opened 2022-06-09
[WIP] Show return_to_sp_url field for all protocols in service provider forms

julialeague:
I noticed something was up when I was [examining a service provider configuration](https://dashboard.int.identitysandbox.gov/service_providers/1699) and saw that the `return_to_sp_url` was populated in the yaml block at the bottom of the page, but I saw no "Return to app url" field showing on the same page. If that attribute is set, I would expect the field to show up.

### Description of Changes:
My understanding of the `return_to_sp_url` attribute is that it can be set with any protocol - looking at idp code confirms that for me. And it makes sense, since it's more of a UX setting than anything protocol-related. 

There were changes introduced ([1] [2]) in November 2021 that identified `return_to_sp_url` as a SAML-specific attribute, which isn't the case. There was also updates to the form to make `return_to_sp_url` a required field for the SAML protocol only (as far as I know, that was the intention).

I updated the service_provider form and show page HTML to declassify this field as a saml-field, so that it shows up no matter which protocol is selected. I also updated the event-listener toggle functions for the protocol radio buttons, so that the form only requires the field to be filled in on submit for the SAML protocol configurations.

Big caveat to all of this is that I could be wrong about my assumptions for how this should work.

[1]: https://github.com/18F/identity-dashboard/commit/1a2f7c95b6b5bc160e0fea4d418e3fbb7deb1b4c#diff-99ef76b66faaa43293c7866031bd1936cf873fdee66d973c6cb8cfafa752479fR84-L95
[2]: https://github.com/18F/identity-dashboard/pull/485

### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/app/javascript/app/service-provider-form.js
+++ b/app/javascript/app/service-provider-form.js
@@ -1,6 +1,6 @@
 const showElement = (element) => element.classList.remove('display-none');
 const hideElement = (element) => element.classList.add('display-none');
-
+  /* eslint no-console: ["error", { allow: ["warn", "error", "log"] }] */
 function ialOptionSetup() {
   // Selectors
   const ialLevel = document.getElementById('service_provider_ial');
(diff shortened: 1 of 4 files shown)


app/javascript/app/service-provider-form.js
```diff
@@ -54,28 +54,51 @@ function protocolOptionSetup() {
...
+    returnToSpUrl.removeAttribute('required');
+    returnToSpUrlLabel.classList.remove('usa-input-required', 'required');
+    returnToSpUrlLabel.removeChild(returnToSpUrlLabel.lastElementChild); // this removes the abbr element that contains the "required" text from the field
+  };
+
+  const addRequiredReturnUrlValidation = () => {
+    returnToSpUrl.setAttribute('required', 'required');
+    returnToSpUrlLabel.classList.add('usa-input-required', 'required');
+    abbrElement.setAttribute("class", "text-secondary-dark");
+    abbrElement.appendChild(requiredText);
```
zachmargolis, line 74: IMO I think it's easier to keep track of changes across JS and ERB if we have everything in the DOM already but toggle attributes on them, like required, hidden, etc etc\

Is it possible to add these attributes in the ERB and hide them?

spec/features/service_providers_spec.rb
```diff
@@ -257,7 +257,7 @@
       choose('service_provider_identity_protocol_openid_connect_private_key_jwt', allow_label_click: true)
 
       saml_attributes =
-        %w[acs_url assertion_consumer_logout_service_url sp_initiated_login_url return_to_sp_url block_encryption signed_response_message_requested]
```
zachmargolis, line 260: should we add some sort of positive test that checks that the field is on the page in both cases?

julialeague:
As I half-suspected, it turns out that hiding this attribute from OIDC configurations is intentional. There is conversation around using a different field to represent the list of connected applications or using this attribute, but no decision has been made yet. Closing this indefinitely.

PR #559, opened 2022-06-09
Update link to be button to fix sign-in

mitchellhenke:

It looks like the jQuery removed in #556 was taking care of some magic of turning links into form submissions, including the sign in to the application.  Relevant slack thread is [here](https://gsa-tts.slack.com/archives/C01SU3NB9T3/p1654812503795529).  This PR changes the link to a button to ensure it submits a POST rather than a GET.

I suspect this same issue could affect https://github.com/18F/identity-dashboard/blob/b2451fb760dcdb64b625f98c4b53a7e8d35d657b/app/views/users/index.html.erb#L38, https://github.com/18F/identity-dashboard/blob/b2451fb760dcdb64b625f98c4b53a7e8d35d657b/app/views/users/edit.html.erb#L10, and https://github.com/18F/identity-dashboard/blob/b2451fb760dcdb64b625f98c4b53a7e8d35d657b/app/views/teams/edit.html.erb#L26

| Old | New |
|----|------|
|  image | image |

### Description of Changes:
  | I switched the link to be a button to ensure it submits a POST.
### PR Checklist:

### PR Review Standards: 


app/views/layouts/application.html.erb
```diff
@@ -136,7 +136,7 @@
               </li>
               <% else %>
               <li class="usa-nav__primary-item">
-                <%= link_to t('links.sign_in'), '/auth/logindotgov', method: :post, class: 'usa-nav__link usa-current' %>
+                <%= button_to t('links.sign_in'), '/auth/logindotgov', method: :post, class: 'usa-button usa-current' %>
```
zachmargolis, line 139: I am 90% certain we have USWDS styles that can format as a plain link? but i don't feel strongly

zachmargolis (approved):
LGTM

app/views/layouts/application.html.erb
mitchellhenke: If we do, that'd be great!  I need something similar for some IDP work.

zachmargolis: Confirmed: [`.usa-button--unstyled`](https://designsystem.digital.gov/components/button/) , [we use it a bunch in the IDP](https://github.com/18F/identity-idp/search?q=usa-button--unstyled)

jmhooper (approved):

app/javascript/packs/application.js
```diff
@@ -0,0 +1 @@
+import '@rails/ujs'
```
zachmargolis, line 1: I would vote to see if we can do `button_to` everywhere and not do JS

mitchellhenke: I added a commit in 514f5e4 to fix the links to be buttons where needed, but the other tough part is we've also lost the confirm dialogs on submission/click (ex: https://github.com/18F/identity-dashboard/blob/77711675f22b15616c641b6fd7fdc59ffdf9f9a3/app/views/users/edit.html.erb#L11)

jmhooper: Can we live without the confirmation dialogues? If so happy to drop UJS, but also kinda want to move on to the js-bundling setup

zachmargolis (approved):
LGTM

app/javascript/packs/application.js
orenyk: My $0.02 is that we should keep UJS for now to avoid any unexpected regressions - I'm not sure where we're using confirmations or non-GET links elsewhere and it doesn't seem worth it to try and find them by hand.

NatashaJIbrahim:
@jmhooper I don't _think_ this is a dealbreaker but I think we should check with @cdelivuk since this is the one part of the Dashboard that's been updated stylistically. I know I've hit that multiple times and it's always thoroughly annoying.

cdelivuk:
thank you @NatashaJIbrahim. [this is more in line with our account page](https://www.figma.com/file/pjGy4zbHO1JqLipF1VzSzJ/Account-page-(Approved-%7C-Internal)?node-id=115%3A3819), placing the main cta above the delete one. the only difference is that cancel is not as large and is a text link below as well. but, i think this is an improvement and agree that delete coming first in the row can be confusing for users.

PR #561, opened 2022-06-10
Introduce JSBundling

jmhooper:
This commit introduces JSBundling in favor of @rails/webpacker.  Managing and upgrading the @rails/webpacker config has become laborious. Hopefully this changes means we'll have a more explicit and simple config for transpiling JS going forward.


webpack.config.js
```diff
@@ -0,0 +1,35 @@
...
+
+module.exports = {
+  mode: isProductionEnv ? 'production' : 'development',
+  devtool: "source-map",
+  entry: {
+    application: "./app/javascript/packs/application.js",
+    manage_users: "./app/javascript/packs/manage_users.js",
+    service_provider_form: "./app/javascript/packs/service_provider_form.js",
+    validate_logo_size: "./app/javascript/packs/validate_logo_size.js",
+  },
```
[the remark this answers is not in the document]
jmhooper: Yep, was planning on working in something like what the idp has.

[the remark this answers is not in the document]
jmhooper: I ended up copying the logic from the IdP. It look a little footwork to tell webpack to look in the right place.

.circleci/config.yml
```diff
@@ -54,6 +54,7 @@ jobs:
       - run:
           name: Test Setup
           command: |
+            yarn build
```
aduth, line 57: Do we have feature specs in this repository? I'd wonder if we should be running `rake assets:precompile` (including Sass compilation) as well / instead.

jmhooper: Yeah, we do. I can swap that one in

config/initializers/secure_headers.rb
```diff
@@ -9,6 +10,8 @@
   form_action << %w[localhost:3000] if Rails.env.development?
   connect_src = ["'self'"]
   connect_src << %w[ws://localhost:3036 http://localhost:3036] if Rails.env.development?
+  script_src = ["'self'", '*.newrelic.com', '*.nr-data.net']
+  script_src << "'unsafe-eval'" if Rails.env.development? || Rails.env.test?
```
aduth, line 14: As mentioned in Slack, if we're not using `eval-sourcemap` Webpack devtool, hopefully we can avoid `unsafe-eval` in the CSP.

package.json
```diff
@@ -5,30 +5,34 @@
...
   "dependencies": {
-    "@rails/webpacker": "^4.3.0",
+    "@rails/ujs": "^6.0.5",
+    "fast-glob": "^3.2.7",
     "identity-style-guide": "^2.2.3",
-    "jquery": "^3.6.0",
-    "jquery-ujs": "^1.2.3",
     "kind-of": "^6.0.3",
     "minimist": "^1.2.6",
     "serialize-javascript": "^3.1.0",
```
aduth, line 18: I don't see that we use these dependencies? I wonder if there was some compat issue in the past with Webpacker where we added them? May be safe to remove now.

Gemfile
```diff
@@ -14,7 +14,8 @@ gem 'faraday'
 gem 'identity-hostdata', git: 'https://github.com/18F/identity-hostdata.git', tag: 'v3.3.0'
 gem 'identity-logging', git: 'https://github.com/18f/identity-logging.git', tag: 'v0.1.0'
 gem 'identity_validations', git: 'https://github.com/18f/identity-validations.git', tag: 'v0.7.4'
-gem 'jquery-rails', '>= 4.4.0'
+gem 'jquery-rails', '>= 4.5.0'
```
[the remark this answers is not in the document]
jmhooper: yeah, i'm gonna have to go through and clean this one up once I hopefully clean up the diff by merging that PR

app/views/layouts/application.html.erb
```diff
@@ -136,7 +136,7 @@
               </li>
               <% else %>
               <li class="usa-nav__primary-item">
-                <%= link_to t('links.sign_in'), '/auth/logindotgov', method: :post, class: 'usa-nav__link usa-current' %>
+                <%= button_to t('links.sign_in'), '/auth/logindotgov', method: :post, class: 'usa-button usa-current' %>
```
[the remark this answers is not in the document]
jmhooper: We don't need it, but also it doesn't hurt anything. Again, just gonna go with whatever gets merged on the other PR

package.json
jmhooper: Let me pull them out and we can see what goes horribly wrong.

config/initializers/secure_headers.rb
jmhooper: good catch

babel.config.js
```diff
@@ -1,72 +1,26 @@
...
-        }
-      ],
       [
-        '@babel/plugin-proposal-object-rest-spread',
+        'polyfill-corejs3',
         {
-          useBuiltIns: true
-        }
+          method: 'usage-global',
+          targets: targets ?? '> 1% and supports es6-module',
```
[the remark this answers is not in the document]
jmhooper: oh yeah, whoops

package.json
jmhooper: Nothing went horribly wrong as far as I can tell.

Gemfile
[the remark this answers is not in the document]
jmhooper: Okay, this should be cleaned up since I merged everything in and dealt with conflicts

package.json
```diff
@@ -5,29 +5,30 @@
...
   "dependencies": {
     "@rails/ujs": "^6.0.5",
-    "@rails/webpacker": "^4.3.0",
+    "fast-glob": "^3.2.7",
     "identity-style-guide": "^2.2.3",
-    "kind-of": "^6.0.3",
-    "minimist": "^1.2.6",
-    "serialize-javascript": "^3.1.0",
-    "webpack-merge": "^4.2.2"
+    "webpack": "^5.73.0",
```
aduth, line 16: Do we need this to be a `dependencies` because we're only installing production dependencies in build environments?

- If yes, then we should also be sure to pull anything else required for build into `dependencies` (Babel, core-js)
- If no, then should we move this back to `devDependencies` ?

aduth (approved):
LGTM! So much red 😍

package.json
aduth: FWIW, running `NODE_ENV=production rake assets:precompile` looks to result in an error because `jsbundling-rails` runs `yarn install` and prunes non-production dependencies.

```
$ webpack
assets by status 0 bytes [cached] 1 asset

ERROR in application
Module not found: Error: Can't resolve 'babel-loader' in '/Users/andrewmduthie/Documents/Code/identity-dashboard'
```

Dunno if that matches the expected behavior we'd have in deployed environments (can't confirm locally since `RAILS_ENV=production` is less than cooperatively locally).

aduth:
Maybe not a blocker, but sourcemaps don't appear to be working in my testing. In the developer tools console I see some warnings:

```
DevTools failed to load source map: Could not load content for http://localhost:3001/assets/application.js-a31f6c9dff05e26a926c889460e9dc7be9e0c46e2096085fa3178d034bc17ab0.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE
```

Looks like it's looking for sourcemaps in the form of `[name].js-[hash].map`, rather than what you have configured:

https://github.com/18F/identity-dashboard/blob/25750ad80a9c2af26a1bc5a2d970f00a355e454d/webpack.config.js#L17

package.json
jmhooper: Yep, good catch. I've got a commit to move everything we need to build into dependencies.

NatashaJIbrahim:
👀 This is so exciting 🙌 Thank you all for working on this and making it happen! Major Kudos and gratitude 🙇‍♀️

PR #568, opened 2022-06-30
[Snyk] Upgrade @babel/register from 7.9.0 to 7.17.7

pauldoomgov:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to upgrade @babel/register from 7.9.0 to 7.17.7.</h3>

:information_source: Keep your dependencies up-to-date. This makes it easier to fix existing vulnerabilities and to more quickly identify and fix newly disclosed vulnerabilities when they affect your project.
<hr/>

- The recommended version is **22 versions** ahead of your current version.
- The recommended version was released **4 months ago**, on 2022-03-14.

<hr/>

**Note:** *You are seeing this because you or someone else with access to this repository has authorized Snyk to open upgrade PRs.*

For more information:  

🧐 [View latest project report](https://app.snyk.io/org/18f-wyj/project/3aa99463-3e36-455c-b709-b01d749db513?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;upgrade-pr)

🛠 [Adjust upgrade PR settings](https://app.snyk.io/org/18f-wyj/project/3aa99463-3e36-455c-b709-b01d749db513/settings/integration?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;upgrade-pr)

🔕 [Ignore this dependency or unsubscribe from future upgrade PRs](https://app.snyk.io/org/18f-wyj/project/3aa99463-3e36-455c-b709-b01d749db513/settings/integration?pkg&#x3D;@babel/register&amp;utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;upgrade-pr#auto-dep-upgrades)

PR #569, opened 2022-06-30
Upgrade JavaScript dependencies to latest

aduth:
**Why**: To avoid a slow trickle of security advisory upgrades (#568, #566, #564, #563, #562), upgrade everything to the latest version.

Upgraded using [npm-check-updates](https://www.npmjs.com/package/npm-check-updates):

```
 @babel/core                  ^7.18.2  →  ^7.18.6     
 @babel/preset-env            ^7.18.2  →  ^7.18.6     
 @babel/register               ^7.9.0  →  ^7.18.6     
 fast-glob                     ^3.2.7  →  ^3.2.11     
 identity-style-guide          ^2.2.3  →   ^6.5.0     
 webpack-cli                   ^4.9.2  →  ^4.10.0     
 @18f/eslint-plugin-identity   ^1.0.0  →   ^2.0.0     
 chai                          ^4.2.0  →   ^4.3.6     
 eslint                       ^7.32.0  →  ^8.18.0     
 jsdom                        ^16.5.0  →  ^20.0.0     
 mocha                         ^7.1.1  →  ^10.0.0     
 sinon                         ^9.0.2  →  ^14.0.0  
```

This does _not_ include an upgrade to `identity-style-guide`, which is a bit more involved.

These updates should be relatively safe, since breaking changes only apply for development-related dependencies.

This also moves `@babel/register` to be a `devDependencies`, since it is only used for the `yarn test` test script.


.circleci/config.yml
```diff
@@ -33,6 +37,16 @@ jobs:
       - restore_cache:
           key: identity-dashboard-{{ checksum "Gemfile.lock" }}
 
+      - run:
```
aduth, line 40: These changes are necessary because the container we use for CircleCI appears to be pinned to an older version of Node.js not supported by the version of `eslint-config-airbnb` upgraded through these changes:

```
error eslint-config-airbnb@19.0.4: The engine "node" is incompatible with this module. Expected version "^10.12.0 || ^12.22.0 || ^14.17.0 || >=16.0.0". Got "14.16.0"
```

The changes here will install the latest v14 version (v14.19.3 as of writing).

This is also the same as what we have in the IdP, and matches the version of Node in the build environments.

- https://github.com/18F/identity-idp/blob/main/.nvmrc
- https://github.com/18F/identity-idp/blob/222d56af/.circleci/config.yml#L42-L54

zachmargolis (approved):
LGTM

PR #572, opened 2022-07-12
Update Service Providers page content (Level of Service + AAL options)

mmagsa:
JIRA issue [LG-6613](https://cm-jira.usa.gov/browse/LG-6613)
[login.gov Dashboard language updates - May 2022](https://docs.google.com/document/d/1R55WCfYScz4rm5g0SdIWo2mZQFyLTxGvkgX5CjNPwKs/edit)
### Description of Changes:
Modified Service Providers page sections "Level of Service" and "Default AAL Level" language as per above document:
#### Before
before1
before2
#### After
after1
after2
### PR Checklist:

2. [X] Have you tagged the appropriate dev(s) for review?

### PR Review Standards: 

### DIFF
--- a/spec/features/service_providers_spec.rb
+++ b/spec/features/service_providers_spec.rb
@@ -39,7 +39,7 @@ feature 'Service Providers CRUD' do
       expect(page).to have_content('urn:gov:gsa:openidconnect.profiles:sp:sso:GSA:app-prod')
       expect(page).to have_content('email')
       expect(page).to have_content(user.teams[0].agency.name)
-      expect(page).to have_content('IAL2')
+      expect(page).to have_content(I18n.t('service_provider_form.ial_option_2'))
       expect(page).to have_content(I18n.t('service_provider_form.aal_option_2'))
     end
 


NatashaJIbrahim:
@cdelivuk ya - frustratingly this content went through Content Review already :-/ @orenyk I know this was pushed through as being a bit urgent but I'm wondering if the workflow for content changes in the future should be through @cdelivuk so that we can make sure they don't add to our accessibility issues? 

I say we add it to our growing backlog in JIRA.

app/views/service_providers/_form.html.erb
```diff
@@ -78,9 +78,9 @@
   <%= form.input :default_aal,
                  as: :select,
                  collection: [
-                   ['', nil],
-                   ['AAL2', 2],
-                   ['AAL2 + Phishing-Resistant MFA', 3]],
+                   ['MFA required + remember device up to 30 days (AAL1)', 1],
```
orenyk, line 81: **issue:** I don't think we should change the value to `1` here - with the way the IdP works, `nil` represents our default behavior. Funnily enough we do technically support a value of `1` but we never use it (since it wasn't available in the dropdown) - AAL1 is technically weaker than our default since it doesn't include MFA.

orenyk (changes requested):
Looks good! A few comments about a) removing some unnecessary content and b) maybe cleaning up the duplicate strings using I18n. Let me know if you have any questions!

orenyk:
@NatashaJIbrahim @cdelivuk I just saw your comments - totally agree that we should run these things through @cdelivuk moving forward! We can also definitely propose changes and get them through content approvals if we have usability concerns with the current length.

spec/features/service_providers_spec.rb
```diff
@@ -39,7 +40,7 @@
       expect(page).to have_content('email')
       expect(page).to have_content(user.teams[0].agency.name)
       expect(page).to have_content('IAL2')
```
orenyk, line 42: **comment:** I think this assertion is passing because we say something like "we don't meet IAL2 at this time" in the help text for the level of service question, but it's not accurately checking that the correct option is selected. Do you want to update this assertion to also check for the correct `I18n` content and make sure that still passes?

orenyk (approved):
Looks great! One small comment that jumped on me but this is good to merge once that is resolved.

PR #582, opened 2022-09-28
Remove user from team page

nprimak:

[JIRA ticket](https://cm-jira.usa.gov/browse/LG-7379?filter=-1)
[Figma](https://www.figma.com/file/ZowuqSNkFiIYPPLGOPP9Rk/Partner-Dashboard-Edits--Internal-Admin-User-LG-5756?node-id=42%3A1826)

### Description of Changes:
This is part of a larger epic where we are making the interface for managing users more friendly in the partners dashboard. This one creates a remove confirmation page so users understand what happens when they remove someone from a team. There is no button or link in the app going to this page so it is essentially invisible unless you know the URL. (/teams/:team_id.users/:id/remove_confirm) 

Received confirmation of language in delete confirmation page (it had to be changed from the original, which said "delete user") 

Screen Shot 2022-10-04 at 2 53 49 PM

### PR Checklist:

### PR Review Standards: 


config/locales/en.yml
```diff
@@ -79,6 +83,7 @@ en:
       the apps first.
     user_already_exists: This user already exists
     user_deleted: Success! You have deleted %{email}
+    user_team_deleted: Success! You have removed ${email} from the team.
```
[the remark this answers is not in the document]
nprimak: Yes, thank you!

app/controllers/teams/users_controller.rb
```diff
@@ -0,0 +1,37 @@
+class Teams::UsersController < AuthenticatedController
+   before_action -> { authorize team, policy_class: TeamUsersPolicy }
+
+    def remove_confirm
+      @team = team
```
julialeague, line 5: I suspect we don't actually need to call team here, since we should be calling it before every action, so the instance variable is already set. I vote we remove this line if everything still works without it.

app/controllers/teams/users_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      @user = team.users.find_by(id: params[:id])
```
julialeague, line 6: **change suggested** We could make this another method similar to the team method below since we do the same thing in two methods.

app/controllers/teams/users_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      if @user.present? && @user != current_user
```
julialeague, line 7: Similarly to the above, we could pull this logic out into a method, since we're using the same logic again below in 'destroy'

app/controllers/teams/users_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+        render_401
+      end
+    end
+  
+    def destroy
+      user = team.users.find_by(id: params[:id])
+      if user.present? && user != current_user
+        team.users.delete(user)
+        flash[:success] = I18n.t('team.users.remove_success', email: user.email)
+        redirect_to team_path(@team.id)
```
julialeague, line 19: You can change the reference to team here as just "team" with no "@", since the team method below is taking care of "memoizing" (i.e. caching) the team.

app/controllers/teams/users_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      else
+        render_401
+      end
+    end
+  
+    def none; end
```
julialeague, line 25: I suspect this can be removed - I don't think it's being used anywhere

app/controllers/teams/users_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  
+    private
+  
+    def user_params
```
julialeague, line 29: This isn't being used, is it?

config/locales/en.yml
```diff
@@ -34,7 +34,16 @@ en:
     users:
       edit: Edit user
       index: Users
+      delete: Delete user
```
julialeague, line 37: It looks like we don't need to add this if it didn't exist previously

config/routes.rb
```diff
@@ -16,6 +16,13 @@
     resources :manage_users, only: %i[new create]
   end
 
+  scope module: 'teams' do
```
julialeague, line 19: After looking into the current routes and controllers a bit, I think what we actually want to do here is add the users resource to the teams resource above, in the same nesting where the 'manage_users' resource is located. This block is essentially the same. I can explain more if you'd like!

nprimak: I think I understand .. I'll take a stab at it now!

julialeague: Never mind, after realizing that we'd have to explicitly point to the teams/users_controller if we changed it as I suggested above, I understand why this is preferable. The scope module: 'teams' line is what's keeping us from explicitly pointing to that controller.

spec/controllers/teams/users_controller_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+    end
+
+    context 'when user not an admin but is a team member and user to delete is team member' do
+        before do
+            team.users << user_to_delete
+            team.users << user
+            delete :destroy, params: { team_id: team.id, id: user_to_delete.id }
+        end
+
+        it 'has a redirect response' do
```
julialeague, line 82: Would you add a test to see that the user was removed from the team as expected?

spec/controllers/teams/users_controller_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+describe Teams::UsersController do
+  include Devise::Test::ControllerHelpers
+
+  let(:user) { create(:user) }
+  let(:team) { create(:team) }
+  let(:user_to_delete) { create(:user) }
+
+  before do
+    allow(controller).to receive(:current_user).and_return(user)
+    sign_in user
```
julialeague, line 12: Could you also add a couple of tests for a signed-out user? I just want something confirming that we're checking that the user is authenticated before anything else. since it's a common concern, I'm not sure if we have any shared examples or context already written in the specs.

spec/controllers/teams/users_controller_spec.rb
```diff
@@ -84,6 +96,19 @@
...
     end
 
+    context 'when user is a team member and user to delete is team member' do
+      before do
+          team.users << user_to_delete
+          team.users << user
+          delete :destroy, params: { team_id: team.id, id: user_to_delete.id }
+      end
+
+      it 'user no longer in team' do
```
julialeague, line 106: Beautiful spec! Only thing I suggest you change is to move it all into the same context as the context starting at line 87. The nice thing about contexts is that you can share them between tests when you're setting up a similar state!

julialeague (approved):
🥳

PR #584, opened 2022-10-20
Use regex for checking email domains

NatashaJIbrahim:
### Description of Changes:

- Regex will allow us to accept all `state.*.us` email addresses. These domains are earmarked for states and are official government domains that we should allow.

- Also improves the description for the method that checks the domain of a prospective user's email address.

### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/app/helpers/team_helper.rb
+++ b/app/helpers/team_helper.rb
@@ -1,19 +1,19 @@
 module TeamHelper
-  WHITELISTED_DOMAINS = %w[.mil .gov .fed.us].freeze
+  ALLOWED_DOMAINS = /@*.mil|@*.gov|@*.fed.us|@*state.*.us/
 
   def can_edit_teams?(user)
     !user.teams.empty? || user.admin?
   end
 
   def can_create_teams?(user)
-    whitelisted_user? || user.admin?
+    allowed_email? || user.admin?
   end
 
   def can_delete_team?(user)
     user.admin?
   end
 
-  def whitelisted_user?
-    WHITELISTED_DOMAINS.any? { |domain| current_user.email.end_with? domain }
+  def allowed_email?
+    ALLOWED_DOMAINS.match?(current_user.email)
   end
 end
--- a/app/helpers/team_helper.rb
+++ b/app/helpers/team_helper.rb
@@ -1,5 +1,5 @@
 module TeamHelper
-  ALLOWED_DOMAINS = /@*.mil|@*.gov|@*.fed.us|@*state.*.us/
+  ALLOWED_DOMAINS = /@.+\.(mil|gov|fed\.us|state\..+\.us)$/
 
   def can_edit_teams?(user)
     !user.teams.empty? || user.admin?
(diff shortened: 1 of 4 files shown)


app/helpers/team_helper.rb
```diff
@@ -1,19 +1,19 @@
 module TeamHelper
-  WHITELISTED_DOMAINS = %w[.mil .gov .fed.us].freeze
+  ALLOWED_DOMAINS = /@*.mil|@*.gov|@*.fed.us|@*state.*.us/
```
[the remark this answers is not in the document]
NatashaJIbrahim: ah! I'm surprised this was working in Rubular then. I'll update it. Thank you!

spec/helpers/team_helper_spec.rb
```diff
@@ -20,4 +22,10 @@
       expect(can_edit_teams?(user)).to eq(true)
     end
   end
+
+  describe '#allowed_email?' do
+    it 'allows expected emails' do
+      expect(allowed_email?).to be false
+    end
+  end
```
aduth, line 30: Could we have specs for all of the domain variants we're trying to support? Or at least an example where `allowed_email?` would be `true` ?

app/helpers/team_helper.rb
```diff
@@ -1,19 +1,19 @@
 module TeamHelper
-  WHITELISTED_DOMAINS = %w[.mil .gov .fed.us].freeze
+  ALLOWED_DOMAINS = /@.+\.(mil|gov|fed\.us|state\..+\.us)$/
```
zachmargolis, line 2: I think this would be a good spot to use the `mail` gem to extract the domain and only check that part, rather than keep that as part of the regex, [people can get sneaky and add multiple @'s inside an address](https://stackoverflow.com/a/12355882)

```ruby
Mail::Address.new('"foobar@bar.baa"@bar.com').domain
# => "bar.com"
```

NatashaJIbrahim: I totally agree. I'm in a tricky spot because we have a partner that has a `state.*.us` email address and it's limiting their ability to create teams and use the Dashboard as intended. This is sort of a patch. 

Do you think this would be too risky as-is? This still relies on Login for the actual login process so would Login let a user have an email address formatted like that?

spec/helpers/team_helper_spec.rb
NatashaJIbrahim: My only hesitations was that having non `example.com` email addresses in specs kindof makes me uneasy BUT it should be fine here right?

app/helpers/team_helper.rb
zachmargolis: Just to be clear!! I was saying let's drop regexing for `@` and just extract the domain and regex that part

I tried for 5 seconds to break the current regex and failed, I think the anchor `$` saves us from anything egregious, so I wouldn't worry about this for now

NatashaJIbrahim: <3 I think we can also add the team via the console for now and then I can give myself a little more time to work on this and explore using the gem.

zachmargolis:
This got out of date so I'm closing it for now

PR #596, opened 2022-11-30
Add dependabot.yml so we can get auto-updates

zachmargolis:

N/A

### Description of Changes:

I noticed that snyk makes a bunch of PRs to this repo, but dependabot doesn't. Snyk has some weird credentials/permissions (auto-creating on behalf of individuals), so I'm hoping that this is what it takes to get dependabot to check here too?

### PR Checklist:

3. [ x Have you linked to any relevant tickets or conversations?

### PR Review Standards: 

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
```
zachmargolis, line 5: I copied this from https://github.com/18F/identity-idp/blob/main/.github/dependabot.yml, but yanked the `identity-style-guide` bit since I wasn't sure if we wanted to update this as often

mitchellhenke (approved):

aduth:
As-is, I think this is going to create pull requests for every dependency update, not just security updates. Is that what we want?

(The intention for the configuration in `identity-idp` to allowlist `identity-style-guide` was that we wanted the behavior to auto-update on every new version)

aduth:
If what we want is security updates, I think we manage that here: https://github.com/18F/identity-dashboard/settings/security_analysis

Screen Shot 2022-11-30 at 3 00 41 PM

zachmargolis:
> As-is, I think this is going to create pull requests for every dependency update, not just security updates. Is that what we want?

I think you're right! Will close.

PR #597, opened 2022-12-06
Attempt to replace id_token with client_id

Jeremy1026:
changelog: replace id_token with client_id in the logout method

LG-7794

### Description of Changes:

Replaces id_token with client_id in the logout code to match current standards.


Gemfile
```diff
@@ -22,8 +22,9 @@ gem 'jwt'
 gem 'kaminari'
 gem 'newrelic_rpm', '>= 6.14.0'
 gem 'nokogiri', '~> 1.13'
-gem 'omniauth_login_dot_gov', git: 'https://github.com/18f/omniauth_login_dot_gov.git'
-gem 'omniauth-rails_csrf_protection', '~> 0.1', '>= 0.1.2'
+gem 'omniauth_login_dot_gov', :git => 'https://github.com/18f/omniauth_login_dot_gov.git',
+                              :branch => 'main'
```
zachmargolis, line 26: can we bring back the "new-style" hash syntax?
```suggestion
gem 'omniauth_login_dot_gov', git: 'https://github.com/18f/omniauth_login_dot_gov.git',
                              branch: 'main'
```
and maybe pin to a tag instead of a branch?
```suggestion
gem 'omniauth_login_dot_gov', git: 'https://github.com/18f/omniauth_login_dot_gov.git',
                              tag: 'v0.2.0'
```

zachmargolis: PS I just pushed a `v1.0.0` tag since it looks like we didn't make one previously

nprimak (approved):
Tested it out and LGTM! I was curious though, why are there so many gem updates?

PR #581, opened 2022-09-20
Add Add New User page (LG-7376)

mmagsa:
### Relevant Ticket or Conversation:
https://cm-jira.usa.gov/browse/LG-7376
https://www.figma.com/file/ZowuqSNkFiIYPPLGOPP9Rk/Partner-Dashboard-Edits--Internal-Admin-User-LG-5756?node-id=42%3A850
### Description of Changes:

Add Add New User page as part of dashboard redesign.
To test, visit: /teams/<team_id>/users/new
replacing <team_id> with the team's id

### PR Checklist:

1. [X] Have you linted and tested your code locally prior to submission?
2. [ ] Have you tagged the appropriate dev(s) for review?
3. [X] Have you linked to any relevant tickets or conversations?

### PR Review Standards: 
- Consider using [Conventional Comments](https://conventionalcomments.org/) to ensure that your feedback is clear and actionable.
- Ideally, PRs should be reviewed by at least 2 team members.
- All PRs must be approved before being merged.

PR #607, opened 2023-01-09
LG-7376 Create Add New User Page (with more validations + tests) 

nprimak:
 [LG-7376](https://cm-jira.usa.gov/browse/LG-7376) 

### Description of Changes:

Added validation and additional tests to Michael's work on 7376 from [earlier stale PR](https://github.com/18F/identity-dashboard/pull/581). UserTeam model validates that user in a team must be unique. This also broke tests in ManageUserForm which had to be updated to account for new validation. Note that for testing "invalid" string as an invalid email, the browser will automatically reject it because of built in validation. You can override it in developer tools by changing "type=email" to "type=text" 

UPDATE ( 01.17.2023) : 
Merged in Michaels changes from LG-7879 (adding feature spec) 

### PR Checklist:

### PR Review Standards: 


spec/controllers/teams/users_controller_spec.rb
```diff
@@ -12,6 +12,93 @@
...
+    let(:invalid_email) { 'invalid' }
+
+    context 'when the user is not part of the team' do
+      it 'renders an error' do
+        post :create, params: { team_id: team.id, user: { email: user_email } }
+        expect(response.status).to eq(401)
+      end
+    end
+
+    context 'when ther user is part of the team' do
```
julialeague, line 53: nit: 'the"

app/controllers/teams/users_controller.rb
```diff
@@ -1,36 +1,64 @@
...
+  def member_email
+    user_params.require(:email).downcase
+  end
+
+  def new_member
+    @new_member ||= User.find_or_create_by!(email: member_email)
+  end
+
+  def user
+    @user ||= team.users.find_by(id: params[:id])
```
orenyk, line 46: **question:** would it make more sense to use `team_and_users` here? Or, in the interest of brevity, just calling it `team` and running `includes(:users)` all the time? I don't think the eager loading will ever have a real performance impact.

app/models/user.rb
```diff
@@ -9,6 +9,8 @@ class User < ApplicationRecord
   has_many :service_providers, through: :teams
   has_many :security_events, dependent: :destroy
 
+  validates :email, format: { with: Devise.email_regexp }
```
orenyk, line 12: **praise:** nice to see this included!

orenyk (commented):
I left a few comments but got a little stuck when I tried to test this locally - let's chat about this tomorrow!

.nvmrc
```diff
@@ -1 +1 @@
-14
```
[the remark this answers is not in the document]
nprimak: I'm not sure how this change got in here..

app/controllers/teams/users_controller.rb
nprimak: changed it to team_and_users

app/controllers/teams/users_controller.rb
```diff
@@ -1,36 +1,64 @@
...
+  def new
+    @user = User.new
+  end
 
-    def user_present_not_current_user(user)
-      user.present? && user != current_user
+  def create
+    new_member.user_teams << UserTeam.create!(user_id: new_member.id, group_id: team.id)
+    flash[:success] = I18n.t('teams.users.create.success', email: member_email)
+    redirect_to team_path(team.id) and return
```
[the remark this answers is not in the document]
nprimak: yeah I agree it makes more sense to redirect to add more team members

.nvmrc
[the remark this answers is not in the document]
orenyk: I'd recommend removing it from the diff if it's not related. Same deal with the `.gitignore` change unless it's related.

app/controllers/teams/users_controller.rb
```diff
@@ -1,36 +1,64 @@
...
+
+  def user_params
+    params.require(:user).permit(:email)
+  end
+
+  def team_and_users
+    @team ||= Team.includes(:users).find(params[:team_id])
+  end
+
+  def team
```
orenyk, line 61: **suggestion:** I think we should just add `.includes(:users)` here - if we use both `team` and `team_and_users` in the same request we could run into issues where it doesn't do the eager loading since `@team` will already exist. Let me know you feel strongly about keeping both, though!

app/models/user_team.rb
```diff
@@ -5,4 +5,7 @@ class UserTeam < ApplicationRecord
 
   belongs_to :user
   belongs_to :team, foreign_key: 'group_id', inverse_of: :user_teams
+
+  validates_uniqueness_of :user_id, scope: :group_id, on: :create,
```
orenyk, line 9: **comment (non-blocking):** one thought I had is that if someone tries to add a user that's already on the team we don't return an error, just proceed with a warning (since it's not a risky request, just redundant). That said, I don't think we should tackle that in this ticket but consider it for consideration as a future improvement. Let me know what you think!

app/views/teams/users/new.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1 class="usa-display">Add New user</h1>
+
+<%= simple_form_for(@user,
+    html: { autocomplete: 'off', role: 'form' },
+    url: team_users_path(@team)) do |form| %>
+  <%= render 'form', form: form %>
+  <div class="margin-top-2">
+    <%= form.button :submit, 'Add', :class => "usa-button" %>
+    <%= link_to 'Back', team_manage_users_path(@team)+'/new',
```
orenyk, line 9: **comment:** based on conversation with @cdelivuk earlier today we'll probably want to remove float-left so that Add comes first.

app/views/teams/users/new.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1 class="usa-display">Add New user</h1>
```
orenyk, line 1: **suggestion:** can we add the team name here? Otherwise it can be confusing / users have issues remembering the context of which team they're adding for.

app/controllers/teams/users_controller.rb
```diff
@@ -1,36 +1,64 @@
...
 
-    def user_present_not_current_user(user)
-      user.present? && user != current_user
+  def create
+    new_member.user_teams << UserTeam.create!(user_id: new_member.id, group_id: team.id)
+    flash[:success] = I18n.t('teams.users.create.success', email: member_email)
+    redirect_to new_team_user_path and return
+  rescue ActiveRecord::RecordInvalid => e
+    flash[:error] = "Email '#{member_email}': " + e.record.errors.full_messages.join(', ')
+    redirect_to new_team_user_path
```
orenyk, line 14: **suggestion (non-blocking):** after playing with this locally I personally think redirecting back to the manage users page makes more sense, even if means users will have to click an extra time to add additional users. With that said, if we add the name of the team to the new user page, maybe it's not such a big deal? I don't have super strong feelings, defer to @cdelivuk and/or your thoughts!

orenyk (approved):
A few minor comments but overall I think this is great! Let me know if you have any questions, nice work wrapping this all up!

PR #627, opened 2023-03-23
Np/lg 9181/cleanup yaml

nprimak:

 [LG-9181](https://cm-jira.usa.gov/browse/LG-9181)

### Description of Changes:
Cleaning up the existing YAML output in the dashboard for PSE's by removing quotation marks so there is no longer an issue of having to add them in places where they were previously missing. Also moving all URL's to the bottom so they are grouped together. 

### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/app/helpers/service_provider_helper.rb
+++ b/app/helpers/service_provider_helper.rb
@@ -73,8 +73,7 @@ module ServiceProviderHelper
 
   def yamlized_sp(service_provider)
     key_from_issuer = service_provider.issuer
-    yamlable_json = { "#{key_from_issuer}": config_hash(service_provider) }
-    yamlable_json.deep_stringify_keys!
+    yamlable_json = { "#{key_from_issuer}" => config_hash(service_provider) }
     yamlable_json.to_yaml.delete('\"')
   end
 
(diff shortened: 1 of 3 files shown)


app/helpers/service_provider_helper.rb
```diff
@@ -72,9 +72,9 @@ def redirect_uri_list(service_provider = @service_provider)
   end
 
   def yamlized_sp(service_provider)
-    key_from_issuer = JSON.parse(service_provider.to_json).dig('production_issuer').presence ||
-                      service_provider.issuer
-    yamlable_json = { "'#{key_from_issuer}'" => config_hash(service_provider) }
+    key_from_issuer = service_provider.issuer
+    yamlable_json = { "#{key_from_issuer}": config_hash(service_provider) }
+    yamlable_json.deep_stringify_keys!
```
mitchellhenke, line 77: It's definitely not critical, but I think this became necessary with the switch from string keys to symbol keys. Is it still needed if the keys are switched back to strings?

mitchellhenke: Apologies, I might be misunderstanding, but my comment is only for the internal Ruby keys used in the hash rather than the keys or values in the resulting YAML.

nprimak: Yeah I think I misunderstood what you were saying (didn't see your response and deleted my comment because I was going to rewrite) . You are correct that we wouldn't need the `deep_stringify_keys` if we kept it as string keys. I don't have a super strong opinion about it besides it seems like the more popular stylistic choice to go with symbols.

mitchellhenke: I don't feel particularly strongly, though in this case I lean towards string keys as they were previously string keys, and YAML returns string keys from parsing so consistency in that regard can help with interoperability back and forth.

nprimak: Gotcha, in that case I will revert the change!

Jeremy1026 (approved):

Sgtpluck (approved):

PR #630, opened 2023-04-03
[LG-9310] SAML signature check via the Dashboard

mmagsa:
https://cm-jira.usa.gov/browse/LG-9310
### Description of Changes:
This change adds a tool for checking the validity of SAML signatures to the Dashboard (under Tools->SAML)
### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/app/controllers/tools_controller.rb
+++ b/app/controllers/tools_controller.rb
@@ -45,7 +45,7 @@ class ToolsController < ApplicationController
   end
 
   def auth_request
-    @auth_request ||= SamlIdp::Request.from_deflated_request(auth_url, get_params: saml_params)
+    @auth_request ||= SamlIdp::Request.from_deflated_request(saml_params[:SAMLRequest], get_params: saml_params)
   end
 
   def auth_service_provider
@@ -61,9 +61,10 @@ class ToolsController < ApplicationController
   end
 
   def saml_params
-    @saml_params ||=
-      saml_params = url_params(auth_url)
-      saml_params.present? ? saml_params : { SAMLRequest: auth_url }
+    @saml_params ||= begin
+      s_params = url_params(auth_url)
+      s_params.present? ? s_params : { SAMLRequest: auth_url }
+    end
   end
 
   def sp


app/controllers/tools_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+    return if !auth_params
+
+    sp = ServiceProvider.find_by(issuer: auth_request.issuer)
+
+    begin
+      certs = (get_cert_param || sp.certs).map { |cert| OpenSSL::X509::Certificate.new(cert) }
+    rescue OpenSSL::X509::CertificateError
+      flash[:error] = "Something is wrong with the certificate you submitted."
+    rescue NoMethodError
+      flash[:error] = "Could not find any certificates to use. Please add a certificate to your application configuration or paste one below."
```
zachmargolis, line 23: I feel like it would be clearer to check `certs.empty?` rather than rely on a NoMethodError for this particular error message, `NoMethodError` can come from anywhere

app/controllers/tools_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    begin
+      auth_request.service_provider.certs = certs
+    rescue NoMethodError
+    end
+
+    @valid_request = valid_request
+    @valid_signature = valid_signature
+    @matching_cert_sn = matching_cert_sn
+    # binding.pry
```
zachmargolis, line 34: make sure this gets removed
```suggestion
```

app/controllers/tools_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  def get_cert_param
+    return [params['cert']] if params['cert'].length > 0
+  end
+
+  def get_auth_request_params
+    auth_url = params['auth_url']
+    return nil if auth_url.length < 1
+    params = get_url_params( auth_url )
+    params = {:SAMLRequest => auth_url} if params.length < 1
```
zachmargolis, line 72: recommend new-style hash
```suggestion
    params = {SAMLRequest: auth_url} if params.length < 1
```

app/controllers/tools_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+    @auth_request ||= SamlIdp::Request.from_deflated_request(auth_params[:SAMLRequest], get_params:auth_params)
+  end
+
+  def auth_params
+    @auth_params ||= get_auth_request_params
+  end
+
+  def get_cert_param
+    return [params['cert']] if params['cert'].length > 0
+  end
```
zachmargolis, line 66:
```suggestion
  def get_cert_param
    [params['cert']] if params['cert'].present?
  end
```

- Drop the `return` keyword, it's implicit as the last expression
- Instead of `.length > 0` or `.length < 1` in a bunch of places, I'd recommend `.present?`
- I am super meh on `get_` names in ruby, I'd think of something like `cert_from_param` as a method name

app/controllers/tools_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+    rescue OpenSSL::X509::CertificateError
+      flash[:error] = "Something is wrong with the certificate you submitted."
+    rescue NoMethodError
+      flash[:error] = "Could not find any certificates to use. Please add a certificate to your application configuration or paste one below."
+    end
+
+    begin
+      auth_request.service_provider.certs = certs
+    rescue NoMethodError
+    end
```
zachmargolis, line 29: why would this error occur? if it's nil? if so, we could use `&`, the nil-safe operator
```suggestion
    auth_request&.service_provider&.certs = certs
```

app/controllers/tools_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    @valid_request = valid_request
+    @valid_signature = valid_signature
+    @matching_cert_sn = matching_cert_sn
+    # binding.pry
+    if auth_request
+      xml = REXML::Document.new( auth_request.raw_xml )
```
zachmargolis, line 36: extra spaces here, if Rubocop doesn't catch this I'd recommend we update it so it does
```suggestion
      xml = REXML::Document.new(auth_request.raw_xml)
```

app/controllers/tools_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  def matching_cert_sn
+    auth_request&.service_provider&.matching_cert&.serial
+  end
+
+  def auth_request
+    @auth_request ||= SamlIdp::Request.from_deflated_request(auth_params[:SAMLRequest], get_params:auth_params)
+  end
+
+  def auth_params
+    @auth_params ||= auth_request_params
```
Sgtpluck, line 57: i'd recommend pulling auth_request_params into the body of this method!

```ruby
def auth_params
  @auth_params ||= 
    auth_url = params['auth_url'] ...
end
```

app/controllers/tools_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def cert_param
+    [params['cert']] if params['cert'].present?
+  end
+
+  def auth_request_params
+    auth_url = params['auth_url']
+    return nil if auth_url.length < 1
```
Sgtpluck, line 66: [question] this is a string, right? maybe `auth_url.empty?`

app/controllers/tools_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+    params = url_params(auth_url)
```
Sgtpluck, line 67: so in a rails controller, `params` is a built-in method that returns a hash-like object. by reassigning here, you're creating some side-effects that could have some unintended effects.

i'd recommend instead creating a method called `saml_request_params` (or whatever)

```ruby
def auth_url_params
   @auth_url_params ||= params['auth_url']
end

def saml_request_params
   url_params(auth_url_params)
end
```

app/views/layouts/application.html.erb
```diff
@@ -105,6 +105,18 @@
               <li class="usa-nav__primary-item">
                 <%= link_to 'Security Events', security_events_path, class: 'usa-nav__link' %>
               </li>
+
+              <li class="usa-nav__primary-item">
+                <button class="usa-accordion__button usa-nav__link" aria-controls="unique-id-2">
```
nprimak, line 110: nitpick but can we give the id a more meaningful name than unique-id-2? if thereis also a unique-id-1 i would suggest the same there

app/views/tools/index.html.erb
```diff
@@ -0,0 +1,70 @@
...
+            Public Certificate Serial Number:
+            <div>
+                <code><%= @matching_cert_sn.to_s %></code> (int)
+            </div>
+            <div>
+                <code><%= @matching_cert_sn.to_s(16) %></code> (hex)
+            </div>
+        <% end %>
+        <% if !@valid_signature %>
+            Possible reasons reasons:
```
nprimak, line 44: typo :)

app/views/tools/index.html.erb
```diff
@@ -0,0 +1,70 @@
...
+                </li>
+                <li>
+                    <b>Wrong certificate:</b>
+                    <ul>
+                        <li>
+                            If you've pasted a certificate, the certificate may not match the private key used by your application.
+                        </li>
+                        <li>
+                            Otherwise, you need to add the public certificate to your Login.gov application configuration that belongs to the private key used by your application.
+                        </li>
```
nprimak, line 57: I appreciate that you added all the reasons why someone might have an invalid cert!

Sgtpluck (approved):

Jeremy1026 (approved):

PR #646, opened 2023-05-09
Make strings that have HTML to render as HTML

Jeremy1026:
### Description of Changes:

Some of our internalization strings include HTML but weren't being rendered in HTML. This will force Ruby to render the HTML instead of just printing it out to the user.

### Screenshots

Before:
image

After:
image

### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/app/views/service_providers/show.html.erb
+++ b/app/views/service_providers/show.html.erb
@@ -13,14 +13,14 @@
   <label for="name">App name: </label>
   <span class="font-mono-xs margin-left-5"><%= service_provider.app_name %></span>
 </h2>
-<p><i> <%= I18n.t('service_provider_form.app_name') %> </i></p>
+<p><i> <%= sanitize I18n.t('service_provider_form.app_name') %> </i></p>
 <hr>
 
 <h2 class="margin-bottom-0">
   <label for="name">Friendly name:</label>
   <span class="font-mono-xs margin-left-5"><%= service_provider.friendly_name %></span>
 </h2>
-<p><i> <%= I18n.t('service_provider_form.friendly_name') %> </i></p>
+<p><i> <%= sanitize I18n.t('service_provider_form.friendly_name') %> </i></p>
 <hr>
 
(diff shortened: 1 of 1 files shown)


app/views/service_providers/show.html.erb
```diff
@@ -13,14 +13,14 @@
...
-<p><i> <%= I18n.t('service_provider_form.app_name') %> </i></p>
+<p><i> <%= sanitize I18n.t('service_provider_form.app_name') %> </i></p>
 <hr>
 
 <h2 class="margin-bottom-0">
   <label for="name">Friendly name:</label>
   <span class="font-mono-xs margin-left-5"><%= service_provider.friendly_name %></span>
 </h2>
-<p><i> <%= I18n.t('service_provider_form.friendly_name') %> </i></p>
+<p><i> <%= sanitize I18n.t('service_provider_form.friendly_name') %> </i></p>
```
Sgtpluck, line 23: i think i used `.html_safe` in the forms.  do you know if `sanitize` or `.html_safe` is preferred? should we make it consistent?

app/views/service_providers/show.html.erb
```diff
@@ -13,14 +13,14 @@
   <label for="name">App name: </label>
   <span class="font-mono-xs margin-left-5"><%= service_provider.app_name %></span>
 </h2>
-<p><i> <%= I18n.t('service_provider_form.app_name') %> </i></p>
+<p><i> <%= sanitize I18n.t('service_provider_form.app_name') %> </i></p>
```
zachmargolis, line 16: Another approach is when we have HTML inside of i18n'd strings is to suffix the key with `_html` so that I18n will auto-HTML safe it

More docs on it: https://guides.rubyonrails.org/i18n.html#using-safe-html-translations

zachmargolis: The bonus is that the templates don't need to worry about this (aside from changing the key name)

PR #650, opened 2023-05-16
lg-9761 update serializer to include protocol attribute one more time with feeling

Sgtpluck:
the first attempt to add the protocol attribute led to a bug where partners hit errors when they tried to update their config https://github.com/18F/identity-dashboard/pull/649

this time, we only pass the protocol attribute through if the serializer is being called via our show endpoint.

### Description of Changes:
added a :show option to the Serializer

### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/spec/serializers/service_provider_serializer_spec.rb
+++ b/spec/serializers/service_provider_serializer_spec.rb
@@ -35,7 +35,8 @@ RSpec.describe ServiceProviderSerializer do
           eq(service_provider.email_nameid_format_allowed)
         expect(as_json[:signed_response_message_requested]).to \
           eq(service_provider.signed_response_message_requested)
-        expect(as_json[:allow_prompt_login]).to eq(true)
+        expect(as_json[:allow_prompt_login]).to be true
+        expect(as_json[:protocol]).to eq 'oidc'
       end
     end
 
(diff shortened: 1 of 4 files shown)


spec/serializers/service_provider_serializer_spec.rb
```diff
@@ -35,12 +35,24 @@
           eq(service_provider.email_nameid_format_allowed)
         expect(as_json[:signed_response_message_requested]).to \
           eq(service_provider.signed_response_message_requested)
-        expect(as_json[:allow_prompt_login]).to eq(true)
+        expect(as_json[:allow_prompt_login]).to be true
```
nprimak, line 38: Curious about the difference between `be`  and `eq` here

nprimak (approved):

spec/serializers/service_provider_serializer_spec.rb
zachmargolis: `be` is usually used with more fluid matchers, ex you can say `be >= 1` and it would assert `>=` and there's also the `be_` set of magical prefixes like `be_truthy` and `be_anything` (which would try to call `obj.anything?`)

nprimak: that makes sense thanks @zachmargolis  @Sgtpluck  !

PR #652, opened 2023-06-21
LG-10145 Fix bug that deletes users in update path

Sgtpluck:
https://cm-jira.usa.gov/browse/LG-10145

### Description of Changes:
When a non-admin user tried to update team information, it would delete all the users except that team member. This change adds a hidden_field_tag to pass the existing team's user ids through in order to make sure they are all present in an update.

### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -64,15 +64,19 @@ class TeamsController < AuthenticatedController
   end
 
   def team_params
-    params.require(:team).permit(:name, :agency_id, :description, user_ids: [])
+    params.require(:team).permit(:name, :agency_id, :description)
+  end
+
+  def user_params
+    params[:user_ids]&.split(' ') || []
   end
 
   def update_params_with_current_user
     if current_user.admin?
       team_params
     else
-      user_ids = team_params[:user_ids] || []
-      team_params.merge(user_ids: (user_ids << current_user.id.to_s))
+      user_ids = user_params
+      team_params.merge(user_ids: (user_ids << current_user.id.to_s).uniq)
     end
   end
 end
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -67,7 +67,7 @@ class TeamsController < AuthenticatedController
     params.require(:team).permit(:name, :agency_id, :description)
   end
 
-  def user_params
+  def existing_user_ids
     params[:user_ids]&.split(' ') || []
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/teams_controller.rb
```diff
@@ -64,15 +64,19 @@ def team
   end
 
   def team_params
-    params.require(:team).permit(:name, :agency_id, :description, user_ids: [])
+    params.require(:team).permit(:name, :agency_id, :description)
+  end
+
+  def user_params
+    params[:user_ids]&.split(' ') || []
```
Sgtpluck, line 71: this is necessary because the create path also uses `update_params_with_current_user ` another option would be to change that code path; i wasn't sure which was preferred

app/controllers/teams_controller.rb
```diff
@@ -64,15 +64,19 @@ def team
...
   end
 
   def update_params_with_current_user
     if current_user.admin?
       team_params
     else
-      user_ids = team_params[:user_ids] || []
-      team_params.merge(user_ids: (user_ids << current_user.id.to_s))
+      user_ids = user_params
+      team_params.merge(user_ids: (user_ids << current_user.id.to_s).uniq)
```
zachmargolis, line 79: Is this used for removing team members or just for adding?

Sgtpluck: just adding

Sgtpluck: well that is a little misleading. they are added in a separate form, unless the team is being created. if it's being created, then line 79 is how the current_user is added to the team. if the team is being updated, the team members actually already exist, but since it's using the same code path as create, if they are not explicitly passed through, then the user_ids are JUST the current_user, and the existing users get removed as part of the update. 

this is one solution -- as my comment above notes, there is another one, which is altering what the create path uses. (i don't really love either option and am extremely open to feedback about what is preferred!)

Jeremy1026 (approved):

app/controllers/teams_controller.rb
zachmargolis: I feel like it might be a little less error-prone if we had a separate array for the existing users (the one in the template) and also did a non-mutating `+` instead of `<<` but maybe a candidate for a followup PR

so the end result could look like
```
team_params.merge(user_ids: (existing_user_ids + [current_user_id.to_s]).uniq)
```

Sgtpluck: happy to make that change in this pr!

zachmargolis (approved):
LGTM

PR #662, opened 2023-08-15
Send PKCE attribute in service providers API so it is updated in the IDP

mitchellhenke:

The (hopefully) final large change for [SBAR misconfiguration work](https://docs.google.com/document/d/1oi5FyJqzH-yA5nR6mC_AcDhEp-avpLoZTzYQelPoP5M/edit)

### Description of Changes:

Currently, the dashboard does not send the `pkce` attribute via the API when service providers are updated. The result of this is that all IDP environments that use the dashboard application will have service providers where the `pkce` value is always `nil`. This can be confusing and lead to issues when promoted to production though this is mitigated by [this change](https://github.com/18F/identity-idp-config/pull/1709) in the identity-idp-config repository.

- Passes true or false in the `pkce` attribute with a couple of exceptions (drawn from a similar one [here](https://github.com/18F/identity-idp-config/blob/dfd066f/lib/models/service_provider.rb#L299))

### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/app/serializers/service_provider_serializer.rb
+++ b/app/serializers/service_provider_serializer.rb
@@ -24,11 +24,11 @@ class ServiceProviderSerializer < ActiveModel::Serializer
     :updated_at,
     :help_text,
     :allow_prompt_login,
+    :pkce,
   )
 
   # these are attributes that should only be passed to the show
   # endpoint
-  attribute :protocol, if: :show
   attribute :pkce, if: :show
 
   def show
@@ -48,7 +48,11 @@ class ServiceProviderSerializer < ActiveModel::Serializer
   end
 
   def pkce
-    object.openid_connect_pkce?
+    if object.openid_connect_pkce?
+      true
+    else
+      false
+    end
   end
 
   def protocol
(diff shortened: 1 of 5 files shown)


app/serializers/service_provider_serializer.rb
```diff
@@ -48,7 +48,13 @@ def certs
   end
 
   def pkce
-    object.openid_connect_pkce?
+    if ['urn:gov:gsa:sp:sso:frb_stl:scs', 'urn:gov:gsa:sp:sso:frb_stl:ecs'].include?(object.issuer)
```
zachmargolis, line 51: would it be worth putting these in a config instead of hardcoding in source?

mitchellhenke: Yep, good call, I forgot we also have config here!

zachmargolis (approved):

app/serializers/service_provider_serializer.rb
```diff
@@ -48,7 +48,13 @@ def certs
   end
 
   def pkce
-    object.openid_connect_pkce?
+    if IdentityConfig.store.service_providers_with_nil_pkce.include?(object.issuer)
+      nil
```
Sgtpluck, line 52: [suggestion] i'm wondering if this should remain `object.openid_connect_pkce?` since in theory that partner could decide to change it explicitly to pkce

Sgtpluck (approved):
one lil question but looks good besides that!

PR #671, opened 2023-09-14
LG-10585 Make failure_to_proof_url required if IdV configs

Sgtpluck:
https://cm-jira.usa.gov/browse/LG-10585

### Description of Changes:
This change:

* makes the failure_to_proof_url input field required for IdV configurations
failure_to_proof input with required tag

* adds formatting to the  issuer description to make it more readable as well as easier to copy/paste the issuer template

formatted issuer description

### PR Checklist:

3. [ ] Have you tagged the appropriate dev(s) for review?
4. [ ] Have you linked to any relevant tickets or conversations?

### PR Review Standards: 


config/locales/service_providers.en.yml
```diff
@@ -13,7 +13,7 @@ en:
...
     change_issuer: "The issuer cannot be changed, but you can create a new test app with a different issuer."
-    issuer: "This is a unique string to identify the app in the IdP, and is also used as the <code>client_id</code>. We recommend something like the following, replacing <code>agency_name</code> and <code>app_name</code> with your own.<br><i>For OpenID Connect: </i><code class='ml2'>urn:gov:gsa:openidconnect.profiles:sp:sso:agency_name:app_name</code><br><i>For SAML: </i><code class='ml2'>urn:gov:gsa:SAML:2.0.profiles:sp:sso:agency_name:app_name</code>"
+    issuer: "<p>This is a unique string to identify the app in the IdP, and is also used as the <code>client_id</code>. We recommend something like the following, replacing <code>agency_name</code> and <code>app_name</code> with your own.</p><i>For OpenID Connect: <div></i><code class='ml2'>urn:gov:gsa:openidconnect.profiles:sp:sso:agency_name:app_name</code><br><i></div>For SAML: <div></i><code class='ml2'>urn:gov:gsa:SAML:2.0.profiles:sp:sso:agency_name:app_name</code></div>"
```
nprimak, line 16: Can we remove the divs here? I don't think they would be serving a stylistic purpose and they are also in weird spots.

app/views/service_providers/_form.html.erb
```diff
@@ -278,8 +278,11 @@ B7xxt8BVc69rHV15A0qyx77CLSj3tCx2IUXVqRs5mlvA==
   </div>
   <%= form.input :failure_to_proof_url,
                  input_html: { class: 'usa-input' },
+                 label_html: { class: 'usa-input-required'},
```
nprimak, line 281: do we want a conditional on this styling since its only required for IdV?

nprimak (changes requested):
couple minor things

config/locales/service_providers.en.yml
Sgtpluck: they are serving a purpose. it is putting the issuer line on its own line, which allows a user to double click on `urn:gov:gsa:openidconnect.profiles:sp:sso:agency_name:app_name` to more easily copy/paste it.

Sgtpluck: if there is a better way to do this, i am happy to change it!

app/views/service_providers/_form.html.erb
Sgtpluck: hm, i was going to fix this, but looking at it more closely, i think this is just the red visual indicator. since the `failure_to_proof_url` input only shows up when it's required, i think it's fine to leave it as is, especially since it'll complicate the javascript a bit to include it. what do you think?

nprimak: oh I see, that makes sense. disregard!

config/locales/service_providers.en.yml
nprimak: The <p></p> tags I think would make more sense, or you could try <br /> they would make more sense here than <div> - but there is a possibility the <p> tag might add more space than desired.

Sgtpluck: it does add a bit more space, but i think that's fine -- this seems like something we'd like to make more readable
issuer description with more whitespace

Sgtpluck: code updated in https://github.com/18F/identity-dashboard/pull/671/commits/7fdeacdd20f3d19193be27731f6e2a05c6826063

nprimak (approved):
:shipit:

PR #691, opened 2023-10-27
LG-10591 add warning to saml tool

Sgtpluck:
https://cm-jira.usa.gov/browse/LG-10591
https://cm-jira.usa.gov/browse/LG-10941 (this one hasn't been refined, but I effectively did it)

### Description of Changes:
I know I was supposed to just add an info box, but the HTML wasn't in a great place, and one thing led to another, and I just ended up refactoring the whole component and creating a class for all the SAMLRequest logic. (That really shouldn't be in a controller.) It still could be better, but I didn't want to go too wild since it works.

Other big changes:
* Moved the bulk of the logic to a Tools::SAMLRequest class. If we end up making tools to test other types of SAML requests, we can consider adding subclasses or something.
* Moved the valid/invalid details above the form, because that's the most relevant information to the user.
* Fixed the HTML to make it more accessible and in-line with best practices (it probably still needs some work but I was trying not to rewrite it from scratch)

### PR Checklist:

### PR Review Standards: 


app/controllers/tools_controller.rb
```diff
@@ -1,77 +1,32 @@
...
-    @valid_signature = valid_signature
-    @matching_cert_sn = matching_cert_sn
+    @request = Tools::SAMLRequest.new(validation_params)
 
-    if auth_request && @valid_request
-      xml = REXML::Document.new(auth_request.raw_xml)
-      xml.write(@xml = '', 2)
+    if @request.logout_request?
+      flash[:warning] = 'You have passed a logout request. Currently, this tool is for ' +
+                        'Authentication requests only.'
```
nprimak, line 17: Do we want to add a link somewhere to the SAML tool for testing logout requests for now until we have our own tool?

app/views/tools/saml_request.html.erb
```diff
@@ -0,0 +1,69 @@
...
+  <div class="usa-alert <%= @request.valid ? "usa-alert--success" : "usa-alert--error"%>">
+    <div class="usa-alert__body">
+      <h2 class="usa-alert__heading">
+        Request: <%= @request.valid ? "Valid" : "Invalid" %>
+      </h2>
+      <% if !@request.valid %>
+        <p>
+          We could not decode your SAML request. Make sure it is encoded correctly, as described in
+          <a href="https://developers.login.gov/saml/#auth-request" target="_blank">our documentation</a>.<br />
+          Also note that all other URL parameters need to be properly URL-encoded.
```
nprimak, line 15: Maybe a link to a reputable site that describes what URL encoding is?

nprimak (approved):
This is great, thanks for working on this! Left a couple suggestions where we could potentially link to outside resources, but feel free to take them or leave them.

app/views/tools/saml_request.html.erb
Sgtpluck: great idea. do you think wikipedia is ok? that feels very neutral.

app/controllers/tools_controller.rb
Sgtpluck: yes!

app/views/tools/saml_request.html.erb
Sgtpluck: added in https://github.com/18F/identity-dashboard/pull/691/commits/7d98210d75319007316f4f24596c8c6419595d5a

app/controllers/tools_controller.rb
nprimak: thanks!

app/views/tools/saml_request.html.erb
nprimak: yeah wikipedia works, perfect 🎉

Jeremy1026 (approved):

PR #703, opened 2023-12-20
Fix failure to proof hint so the link displays as a link

Jeremy1026:

[LG-11452](https://cm-jira.usa.gov/browse/LG-11452)

### Description of Changes:

Fix the Failure to Proof hint so that the link is clickable, instead of printing out the HTML.

### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/app/views/service_providers/_form.html.erb
+++ b/app/views/service_providers/_form.html.erb
@@ -280,7 +280,7 @@ B7xxt8BVc69rHV15A0qyx77CLSj3tCx2IUXVqRs5mlvA==
                  input_html: { class: 'usa-input' },
                  label_html: { class: 'usa-input-required'},
                  label: 'Failure to Proof URL',
-                 hint: I18n.t('service_provider_form.failure_to_proof_url'),
+                 hint: I18n.t('service_provider_form.failure_to_proof_url').html_safe,
                  required: true %>
 
   <%= form.input :push_notification_url,


app/views/service_providers/_form.html.erb
```diff
@@ -280,7 +280,7 @@ B7xxt8BVc69rHV15A0qyx77CLSj3tCx2IUXVqRs5mlvA==
                  input_html: { class: 'usa-input' },
                  label_html: { class: 'usa-input-required'},
                  label: 'Failure to Proof URL',
-                 hint: I18n.t('service_provider_form.failure_to_proof_url'),
+                 hint: I18n.t('service_provider_form.failure_to_proof_url').html_safe,
```
mitchellhenke, line 283: I think we should prefer renaming the key of the YAML content to `failure_to_proof_url_html`, which should automatically take care of escaping. Even better would be to separate out the HTML/content, but that'd be more involved.

Jeremy1026: Tried changing the name as suggested and dropping the html_safe, and that doesn't seem to do it automatically. I'll dig into it a bit to see whats going on.

mitchellhenke: https://github.com/18F/identity-idp/blob/67df16d/config/locales/idv/en.yml#L264 is an example on the IDP side

nprimak (approved):

app/views/service_providers/_form.html.erb
mitchellhenke: I think it's as straightforward as using `t()`

mitchellhenke: Oh I missed that this was merged

https://github.com/18F/identity-dashboard/compare/mitchellhenke/translation-helper shows the changes needed

PR #714, opened 2024-01-09
Updating pull policy from IfNotPresent to Always to prevent stale images

stephencshelton:
### Description of Changes:
Swaps from the default pull policy of `IfNotPresent` to `Always` for images that are referencing `main` to prevent stale containers from hanging around and being used.

### PR Checklist:

### PR Review Standards: 


timothy-spencer (approved):
This is fine, though thinking about this, shouldn't the tag be changing as we do new releases, and thus not require this change?

stephencshelton:
So the tag is changing but if the review app gets scheduled onto a node that has already pulled the `main` tag already it will think it already has it and won't bother going out to ECR and seeing the SHA that main is now on changed/pulling it again I believe.

PR #744, opened 2024-02-27
Add papertrail admin interface

Jeremy1026:

[LG-12358](https://cm-jira.usa.gov/browse/LG-12358)

### Description of Changes:

Adds a section for version history on the integration configuration page only visible to admins. Shows changes made, grouped by version.

image

### PR Checklist:

### PR Review Standards: 


app/views/service_providers/show.html.erb
```diff
@@ -222,3 +222,44 @@
...
+            <thead>
+              <tr>
+                <th scope="col">Field</th>
+                <th scope="col">Previous Value</th>
+                <th scope="col">Updated Value</th>
+              </tr>
+            </thead>
+          <% v.object_changes.each do |field, change| %>
+            <% if field != 'updated_at' %>
+              <% before, after = change %>
```
zachmargolis, line 249: we can destructure inside the loop
```suggestion
          <% v.object_changes.each do |field, (before, after)| %>
            <% if field != 'updated_at' %>
```

app/views/service_providers/show.html.erb
```diff
@@ -222,3 +222,44 @@
...
+    <button class="usa-accordion__button margin-top-105" aria-expanded="false" aria-controls="versions">
+      <h3 id="versionToggle" class="usa-accordion__heading">Version History (<%= service_provider.versions.count %>) </h3>
+    </button>
+    <div id="versions" class="usa-accordion__content">
+    <% service_provider.versions.reverse.each do |v| %>
+      <div class="event">
+        <h4 class="margin-0">Action: <%= v.event.capitalize %></h4>
+        <h5 class="margin-0">By: <%= v.whodunnit %></h5>
+        <h5 class="margin-0">At: <%= v.created_at %></h5>
+        <% if v.event != 'create' %>
```
Sgtpluck, line 238: [question] is there a reason to leave out the initial create? probably helpful to know the starting point!

Jeremy1026: My thought process there was:

* The create list is rather long
* We can go through changes to get back to the initial values
* If the list of versions is super long making reversing the changes to get to the original difficult, knowing what the original value was probably isn't useful

app/views/service_providers/show.html.erb
Jeremy1026: I like simpler! Committed.

app/views/service_providers/show.html.erb
Sgtpluck: that makes sense. i think there is possibly some utility in knowing the first iteration, but that's just a guess, so we can add it if we need it! thanks 🙏

Sgtpluck (approved):

PR #748, opened 2024-03-01
LG-12330 Update design system to current release

Sgtpluck:
https://cm-jira.usa.gov/browse/LG-12330

### Description of Changes:
This change includes:
- upgrading design system to new naming conventions.
- lots of markup updates to get aligned to uswds and lgds current.
- reworked some html components so that we can remove some custom css
- removed public/assets/uswds directory as it is unnecessary

Note:
The checkboxes/radio buttons spacing is a little bit weird. I wanted to get a PR opened, but can continue to look at those and either update in this PR or another one if necessary

### PR Checklist:

### PR Review Standards: 


mitchellhenke (approved):

app/assets/stylesheets/application.css.scss
```diff
@@ -127,3 +89,9 @@ $image-path: '';
     display: block;
   }
 }
+
+.usa-button--tiny {
```
Sgtpluck, line 93: this button variant was removed from the Login.gov design system, but we don't have an alternative for it currently, so i just ported the css

app/assets/stylesheets/application.css.scss
```diff
@@ -36,45 +37,6 @@ $image-path: '';
   }
 }
 
-.usa-label {
```
Sgtpluck, line 39: i was able to get rid of a lot of our custom css as either things were fixed now, or i was able to use utility classes to achieve the same effect.

public/assets/uswds/css/uswds.css
Sgtpluck, line 1: these assets (at `/public/assets/uswds`) don't need to be checked in!

Jeremy1026 (approved):

aduth:
The active state when clicking a dropdown navigation menu has some pretty several text contrast issues:

image

aduth:
Not sure how easy it'd be to fix, but there's some pretty severe flickering of the menu when the page initially loads.

https://github.com/18F/identity-dashboard/assets/1779930/55500cb7-14cd-4ec3-98ff-5b7e7fe774f4

aduth (approved):
Some cosmetic issues noted, but LGTM overall 👍

aduth:
When trying to click one of the expandable sections, it flips the icon to show as being expanded, but doesn't actually list the sub-menu items?

image

Sgtpluck:
@aduth there is some very strange interaction happening now -- the first time i click the expandable sections, it doesn't show the submenu, like you said. however, if i move the cursor away and then back, it does expand. it also expands/collapses correctly every time after the first. is this header haunted?

app/javascript/packs/uswds-init.js
aduth, line 1: I think we can reference this from `node_modules` rather than including a copy of it.

Can probably delete this and patch in the `@uswds/uswds` load path:

```diff
diff --git a/config/initializers/assets.rb b/config/initializers/assets.rb
index c6e49be..19eeaa9 100644
--- a/config/initializers/assets.rb
+++ b/config/initializers/assets.rb
@@ -5,6 +5,7 @@ Rails.application.config.assets.version = '1.0'
 
 # Add additional assets to the asset load path
 Rails.application.config.assets.paths.push(
+  'node_modules/@uswds/uswds/dist/assets/js',
   'node_modules/@18f/identity-design-system/dist/assets/img',
   'node_modules/@18f/identity-design-system/dist/assets/fonts',
   'node_modules/@18f/identity-design-system/dist/assets/js',
```

Arguably this should probably be included with the Login.gov Design System to be loaded out of the existing `@18f/identity-design-system/dist/assets/js` asset path, but this should work in the meantime.

aduth: Also, we'd probably want to reference `uswds-init.min.js` instead of `uswds-init.js` in the layout if we did that, to continue loading it as minified.

Sgtpluck: i like this! updated in https://github.com/18F/identity-dashboard/pull/748/commits/02c318fa93f0d428a3258330aeb40bc6cac32324

app/assets/stylesheets/application.css.scss
```diff
@@ -1,7 +1,8 @@
-$font-path: '/fonts';
-$image-path: '';
-
-@import 'identity-style-guide/dist/assets/scss/styles';
+@forward 'uswds' with(
+  $theme-font-path: '/fonts',
+  $theme-image-path: '',
+  );
```
zachmargolis, line 4: idk what the typical SASS style is but for any other language I would unindent the last paren:
```suggestion
@forward 'uswds' with(
  $theme-font-path: '/fonts',
  $theme-image-path: '',
);
```

Sgtpluck: updated in https://github.com/18F/identity-dashboard/pull/748/commits/49d6e83970cd1f9fe1118eb7dda8770c84763753

app/assets/stylesheets/application.css.scss
```diff
@@ -127,3 +89,19 @@ $image-path: '';
...
+
+//  TODO: this may be updated in a future Login.gov design system release.
+// remove this if/when it is.
+.usa-nav__primary button {
+  &[aria-expanded='true'] {
+    &:hover {
+      background-color: $theme-color-primary-darker;
+    }
+  }
+}
```
aduth, line 107: Based on [the upstream fix](https://github.com/18F/identity-design-system/pull/419), I suspected this may have an inverse effect of regressing the intended mobile appearance:

image

You can probably use the viewport helper mixin like in the design system pull request to fix it:

```suggestion
//  TODO: this may be updated in a future Login.gov design system release.
// remove this if/when it is.
@include at-media-max($theme-header-min-width) {
  .usa-nav__primary button {
    &[aria-expanded='true'] {
      &:hover {
        background-color: $theme-color-primary-darker;
      }
    }
  }
}
```

Sgtpluck: fixed it to be `at-media($theme-header-min-width) {` and it works! thanks!

aduth:
Minor visual regression, can probably be addressed separately. The buttons on Admin > Users now break onto separate lines.

Before|After
---|---
image|image

aduth (approved):
LGTM 🚀

PR #750, opened 2024-03-07
Dmm/upgrade design system to 9.0

Sgtpluck:
There's a new LGDS!!!!

### Description of Changes:
The only big changes I could identify were:
- Updates the header to the proper colors
- Adds an image that was removed from the design system
- Updates the menu close mobile nav button to an icon
- Adds a helper module to include `usa-current` class when necessary

During the course of this change, I also:
- Created a nav partial to make it easier to work on the navigation
### PR Checklist:

### PR Review Standards: 


app/assets/stylesheets/application.css.scss
```diff
@@ -96,13 +96,15 @@
...
 }
 
-//  TODO: this may be updated in a future Login.gov design system release.
-// remove this if/when it is.
-@include at-media($theme-header-min-width) {
-  .usa-nav__primary button {
-    &[aria-expanded='true'] {
-      &:hover {
-        background-color: $theme-color-primary-darker;
+.usa-header--basic {
```
Sgtpluck, line 99: the LGDS optimizes the --extended header, which we are not currently using.

app/javascript/packs/application.js
```diff
@@ -1,5 +1,4 @@
 import Rails from '@rails/ujs';
-import '@18f/identity-design-system/build/cjs/auto';
```
Sgtpluck, line 2: `main.js` is now browser-optimized, so i could return to using it

app/views/layouts/_nav.html.erb
```diff
@@ -0,0 +1,99 @@
+<div class="usa-overlay"></div>
```
Sgtpluck, line 1: i made a partial because i was tired of scrolling around in the application.html.erb file

app/helpers/application_helper.rb
```diff
@@ -0,0 +1,5 @@
+module ApplicationHelper
+  def classes(route, needed_classes='')
```
aduth, line 2: This helper method would be globally available across all views so wonder if it'd be worth qualifying as being specific to "current page navigation" logic?

e.g.

```suggestion
  def classes_if_current_route(route, needed_classes='')
```

Another idea could be to take advantage of Rails tag helpers support for classes as an array, and filter dynamically.

e.g.

```html
<%= link_to 'Apps', service_providers_path, class: [current_page?(service_providers_path) && 'usa-current', 'usa-nav__link'] %>
```

Or maybe abstracting `link_to`

```suggestion
  def navigation_link_to(text, route)
    link_to text, route, class: [current_page?(route) && 'usa-current', 'usa-nav__link']
```

aduth (approved):
LGTM! 👍 

Only had one optional suggestion / idea.

app/helpers/application_helper.rb
Sgtpluck: hm definitely worth changing. 

i think i like the last option best -- was trying to get the logic out of the view, so i'll update!

Sgtpluck: updated in https://github.com/18F/identity-dashboard/pull/750/commits/eb88d9f8a048032876c240f0196b7d00051eaa84

ajfarkas (commented):
Looks good to me. I'm not yet familiar enough to give a nuanced review, but everything looks on the up and up. Thanks for doing a bit of cleanup as well.

PR #751, opened 2024-03-11
LG-12614 change active column label

ajfarkas:
LG-12614 [Dashboard] Make "active" column more understandable

### Description of Changes:
  - remove col from production apps
  - change Active label in apps list and create to Accessible
  - include title text on Accessible column

### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/app/controllers/service_providers_controller.rb
+++ b/app/controllers/service_providers_controller.rb
@@ -70,11 +70,11 @@ class ServiceProvidersController < AuthenticatedController
 
     @service_providers = [ 
       {
-        type: 'Production Apps',
+        type: 'Production',
         apps: prod_apps,
       },
       {
-        type: 'Sandbox Apps',
+        type: 'Sandbox',
         apps: sandbox_apps,
       },
     ]
@@ -257,11 +257,11 @@ class ServiceProvidersController < AuthenticatedController
   def build_service_provider_array(prod_apps, sandbox_apps)
     return [ 
       {
-        type: 'Production Apps',
+        type: 'Production',
         apps: prod_apps,
       },
              {
-               type: 'Sandbox Apps',
+               type: 'Sandbox',
                apps: sandbox_apps,
              },
     ]
--- a/app/controllers/service_providers_controller.rb
+++ b/app/controllers/service_providers_controller.rb
@@ -12,11 +12,11 @@ class ServiceProvidersController < AuthenticatedController
 
     @service_providers = [ 
       {
-        type: 'My Production Apps',
+        type: 'Production',
         apps: prod_apps,
       },
       {
-        type: 'My Sandbox Apps',
+        type: 'Sandbox',
         apps: sandbox_apps,
       },
     ]
--- a/app/controllers/service_providers_controller.rb
+++ b/app/controllers/service_providers_controller.rb
@@ -10,16 +10,7 @@ class ServiceProvidersController < AuthenticatedController
     prod_apps = all_apps.select { |sp| sp.prod_config == true }
     sandbox_apps = all_apps.select { |sp| sp.prod_config == false }
 
-    @service_providers = [ 
-      {
-        type: 'Production',
-        apps: prod_apps,
-      },
-      {
-        type: 'Sandbox',
-        apps: sandbox_apps,
-      },
-    ]
+    @service_providers = build_service_provider_array(prod_apps, sandbox_apps)
   end
 
   def create
@@ -68,16 +59,7 @@ class ServiceProvidersController < AuthenticatedController
     prod_apps = all_apps.select { |sp| sp.prod_config == true }
     sandbox_apps = all_apps.select { |sp| sp.prod_config == false }
 
-    @service_providers = [ 
-      {
-        type: 'Production',
-        apps: prod_apps,
-      },
-      {
-        type: 'Sandbox',
-        apps: sandbox_apps,
-      },
-    ]
+    @service_providers = build_service_provider_array(prod_apps, sandbox_apps)
   end
 
   def publish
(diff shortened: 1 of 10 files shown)


app/views/service_providers/_service_provider_table.erb
```diff
@@ -3,7 +3,7 @@
 
   <% unless sps[:apps].blank? %>
     <table class="usa-table">
-      <%= render partial: 'service_provider_table_header', locals: { show_created_at: show_created_at } %>
+      <%= render partial: 'service_provider_table_header', locals: { is_prod: sps[:apps][0].prod_config, show_created_at: show_created_at } %>
```
zachmargolis, line 6: I dislike hardcoding array indexes, we can use `.first` here?
```suggestion
      <%= render partial: 'service_provider_table_header', locals: { is_prod: sps[:apps].first.prod_config, show_created_at: show_created_at } %>
```

ajfarkas: Thanks, I'm still getting used to Ruby, and JS doesn't have that method on arrays. Will update.

Sgtpluck: i think rather than using the `sp[:apps]` data  to determine this, it would be preferable to pass the local in, like we're doing with `show_created_at`. IE:

[index view](https://github.com/18F/identity-dashboard/blob/main/app/views/service_providers/index.html.erb#L3)
`<%= render partial: 'service_provider_table', locals: { service_providers: @service_providers, show_created_at: false, is_prod: false } %>`

i am also noticing that the value of `is_prod` and the value of `show_created_at` will be the same, so maybe we should just rename `show_created_at` to `is_prod` and use that to determine if the `created_at` column and the `accessible in the integration environment` column, rather than having two variables to keep track of.

app/views/service_providers/_service_provider_table_header.erb
```diff
@@ -3,7 +3,9 @@
     <th scope="col">Friendly name</th>
     <th scope="col">Issuer</th>
     <th scope="col">IAL</th>
-    <th scope="col">Active</th>
+    <% if !is_prod %>
+      <th scope="col"><abbr title="Accessible in the Integration environment">Accessible</abbr></th>
+    <% end %>
     <th scope="col">Cert Exp</th>
     <% if show_created_at %>
```
Sgtpluck, line 10: with my suggestion in `app/views/service_providers/_service_provider_table.erb`, this would become 

```
<% if is_prod %>
```

Sgtpluck (changes requested):
have a suggestion on how to determine the `is_prod` value -- let me know what you think/any questions

app/views/service_providers/_service_provider_table.erb
Jeremy1026: > so maybe we should just rename `show_created_at` to `is_prod`

Yessss!!! This would be a nice update for future us.

app/views/service_providers/_service_provider_table_header.erb
ajfarkas: For some reason, that's updating the SP Table rows, but not the header.
I don't fully understand why it works at all, if we're statically setting `is_prod` to `false`?

ajfarkas: ah, nevermind: I missed an instance of `show_created_at`. I get how it fits together now.

Sgtpluck: oh you know what, i definitely got confused -- the  `created_at` column is flipped on and off depending on whether it's an admin view or a user view (the admin view, can be found here: [admin all service providers view](https://github.com/18F/identity-dashboard/blob/main/app/views/service_providers/all.html.erb#L3). so this won't work, apologies.

app/views/service_providers/_service_provider_table.erb
Sgtpluck: @Jeremy1026 i think i got confused -- `show_created_at` is admin vs user, not prod vs integration applications. 

@ajfarkas probably the right place to do this would be in the service provider controller, in this array of hashes: https://github.com/18F/identity-dashboard/blob/main/app/controllers/service_providers_controller.rb#L14-L22

app/helpers/service_provider_helper.rb
```diff
@@ -73,9 +73,9 @@ def yamlized_sp(service_provider)
     yamlable_json.to_yaml.delete('\"')
   end
 
-  def sp_active_img_alt(service_provider_is_active)
-    return 'Active service provider' if service_provider_is_active
-    'Inactive service provider'
+  def sp_active_img_alt(service_provider_is_accessible)
```
Sgtpluck, line 76: hm is this method in use? i only see it referenced in tests. i think we can axe this, since we add the alt text directly to the image.

Sgtpluck (approved):
this looks great! could you please remove the method/tests i referenced if it looks like my assessment that it's unused is correct?

Sgtpluck (changes requested):
noted that another one of the methods you found looks unused. either use the suggestion to incorporate the method to reduce duplication or delete it! (there's probably more cleanup that could be done to reduce duplication with this method, but it's okay if that feels out of scope here)

app/controllers/service_providers_controller.rb
```diff
@@ -257,11 +257,11 @@ def body_attributes
   def build_service_provider_array(prod_apps, sandbox_apps)
```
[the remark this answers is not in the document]
ajfarkas: good call. I think I need to switch IDE to something like VSCode that will allow me to do a lookup of the usage of a given function. Sublime is fairly bare-bones, which is great for the purposes I've been using it for, but I think Rails development would be easier with a full-featured IDE.

Sgtpluck (approved):

PR #758, opened 2024-03-20
LG-12748 Require SAML ACSUrl

ajfarkas:
[LG-12748](https://cm-jira.usa.gov/browse/LG-12748)
[Slack Conversation](https://gsa-tts.slack.com/archives/C01SU3NB9T3/p1710941326342639)

### Description of Changes:
Added `(*required)` designation on Assertion Consumer Service URL when partners use SAML auth protocol.  
We are already validating the input to ensure it's a URL string.

### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/app/views/service_providers/_form.html.erb
+++ b/app/views/service_providers/_form.html.erb
@@ -224,7 +224,9 @@ B7xxt8BVc69rHV15A0qyx77CLSj3tCx2IUXVqRs5mlvA==
   <div class='saml-fields'>
     <%= form.input :acs_url,
                    input_html: { class: 'usa-input' },
+                   label_html: { class: 'usa-input-required'},
                    label: 'Assertion Consumer Service URL',
+                   required: true,
                    hint: I18n.t('service_provider_form.assertion_consumer_service_url').html_safe %>
 
     <%= form.input :assertion_consumer_logout_service_url,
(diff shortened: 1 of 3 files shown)


Jeremy1026 (approved):

app/views/service_providers/_form.html.erb
```diff
@@ -224,7 +224,9 @@ B7xxt8BVc69rHV15A0qyx77CLSj3tCx2IUXVqRs5mlvA==
   <div class='saml-fields'>
     <%= form.input :acs_url,
                    input_html: { class: 'usa-input' },
+                   label_html: { class: 'usa-input-required'},
                    label: 'Assertion Consumer Service URL',
+                   required: true,
```
Sgtpluck, line 229: is setting `required` initially an issue if a user doesn't select a different protocol? IE, `toggleFormFields` is never triggered? (or does the function trigger when the page is loaded)

ajfarkas: I will check.

ajfarkas: ok, so I think I would want to remove this `required` because if someone has JS disabled, they would not be able to submit the form, even if they chose OIDC protocol.  
The other `required` fields are universally required.

nprimak: that makes sense to me

ajfarkas:
I'm going to write some tests, so I'll make an update and then re-request a review.

app/views/service_providers/_form.html.erb
ajfarkas: So, that doesn't work with Ruby, and the way we add the `required` bug. I'm going to leave this as is.

nprimak (approved):
LGTM!

PR #761, opened 2024-03-28
LG-11773 new config toggle

ajfarkas:
[Jira](https://cm-jira.usa.gov/browse/LG-11773)
[Slack](https://gsa-tts.slack.com/archives/C01SU3NB9T3/p1710880717210819)

### Description of Changes:
Update Radio button toggle for Production/Sandbox to USWDS tiled radio group, update yes/no text for radios to Production/Sandbox;
Update Accessible radio button toggle to use USWDS tiled radio group, for consistency.

Areas tested/to test: Dashboard > Apps > Create a new app:
  - Dashboard Configuration (formerly Production Configuration) section
  - Accessible section

### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/app/views/service_providers/_form.html.erb
+++ b/app/views/service_providers/_form.html.erb
@@ -5,18 +5,16 @@
   <%= form.label :prod_config,
                  label: 'Dashboard Configuration',
                  class: 'usa-label' %>
-  <ul class="usa-button-group usa-button-group--segmented">
-    <li class="usa-button-group__item">
-      <input type="radio" id="radio-production" name="dashboard-config_radio"
-      value="true" class="usa_radio__input" hidden="true">
-      <label for="radio-production" class="usa-button">Production</label>
+  <%= form.collection_radio_buttons(:prod_config,
+    { "Production" => true, "Sandbox" => false },
+    :last, :first,
+    item_wrapper_tag: :ul,
+    item_wrapper_class: "usa-input-list") do |b| %>
+    <li>
+      <%=  b.radio_button(class: "usa-radio__input usa-radio__input--tile") %>
+      <%=  b.label(class: "usa-radio__label")%>
     </li>
-    <li class="usa-button-group__item">
-      <input type="radio" id="radio-sandbox" name="dashboard-config_radio"
-      value="false" class="usa_radio__input" hidden="true" checked>
-      <label for="radio-sandbox" class="usa-button">Sandbox</label>
-    </li>
-  </ul>
+  <% end %>
 
   <%= form.input :app_name,
                  input_html: { class: 'usa-input' },
--- a/app/views/service_providers/_form.html.erb
+++ b/app/views/service_providers/_form.html.erb
@@ -1,20 +1,22 @@
 <%= form.error_notification %>
 <fieldset class='usa-fieldset'>
 
-  <legend class='usa-sr-only'>Production Configuration</legend>
+  <legend class='usa-sr-only'>Dashboard Configuration</legend>
   <%= form.label :prod_config,
-                 label: 'Production Configuration',
-                 class: 'usa-label',
-                 hint: I18n.t('service_provider_form.prod_config').html_safe %>
-  <%= form.collection_radio_buttons(:prod_config,
-                                    { "Yes" => true, "No" => false },
-                                    :last, :first,
-                                    item_wrapper_tag: :div,
-                                    item_wrapper_class: "usa-radio") do |b| %>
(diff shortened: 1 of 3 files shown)


Jeremy1026:
It might be beneficial to include a screenshot of this kind of change in the future.

Jeremy1026 (approved):

ajfarkas:
Screenshot 2024-03-28 at 10 56 16 AM
Screenshot 2024-03-28 at 10 56 25 AM

app/views/service_providers/_form.html.erb
```diff
@@ -1,20 +1,20 @@
...
+                 class: 'usa-label' %>
   <%= form.collection_radio_buttons(:prod_config,
-                                    { "Yes" => true, "No" => false },
-                                    :last, :first,
-                                    item_wrapper_tag: :div,
-                                    item_wrapper_class: "usa-radio") do |b| %>
-    <%=  b.radio_button(class: "usa-radio__input") %>
-    <%=  b.label(class: "usa-radio__label")%>
+    { "Production" => true, "Sandbox" => false },
+    :last, :first,
```
Sgtpluck, line 10: hm i feel like the formatting difference is wigging me out a little. this is non blocking, but could you make it consistent? (i thought rubocop dealt with this, but maybe not? maybe we should tighten the rules)

app/views/service_providers/_form.html.erb
```diff
@@ -1,20 +1,20 @@
 <%= form.error_notification %>
 <fieldset class='usa-fieldset'>
 
-  <legend class='usa-sr-only'>Production Configuration</legend>
+  <legend class='usa-sr-only'>Dashboard Configuration</legend>
   <%= form.label :prod_config,
-                 label: 'Production Configuration',
-                 class: 'usa-label',
-                 hint: I18n.t('service_provider_form.prod_config').html_safe %>
```
Sgtpluck, line 8: [question] do we still want a hint?

Sgtpluck (approved):
it looks good, but had a couple questions!

app/views/service_providers/_form.html.erb
Sgtpluck: honestly i prefer the way you're indenting, but it would be real out of scope to fix the whole view, so i dunno

nprimak (approved):
Thanks for adding the screenshots!

app/views/service_providers/_form.html.erb
ajfarkas: What does the hint do? I couldn't find any difference bw leaving and removing it.

ajfarkas: I also don't see any documentation for a `hint` attribute with a quick SE search

app/views/service_providers/_form.html.erb
ajfarkas: I can fix that.
  For the record, this indentation makes me very uncomfortable. I'd much rather we do:
```
<% some_function(
  :param1,
  :param2,
  param3: true
) do |var| %>
```

app/views/service_providers/_form.html.erb
Jeremy1026: Hint is the mouse hover alt text I believe.

Jeremy1026: Also accessibility screen reader stuff

ajfarkas:
@Sgtpluck In terms of the hint, do you mean we should keep the visible text that describes the function of the Dashboard Configuration radio/toggle?
Or specifically the Rails `hint` param?

app/views/service_providers/_form.html.erb
Sgtpluck: it looks like the hint SHOULD be getting wrapped with a div with the uswds class `usa-hint` according to this code: https://github.com/18F/identity-dashboard/blob/main/config/initializers/simple_form.rb#L26.

Sgtpluck: ok @ajfarkas and i did a little more investigation -- it looks like we sometimes duplicate the `hint` attribute, both on the `form` and the `label`. 

the `label hint: I18n.t('service_provider_form.prod_config').html_safe` version creates a `hint` label on the attribute which is not a standard attribute and doesn't seem to do anything (and all the information is being rendered in more standard ways via the form anyways), so i believe it's safe to remove. we should also go through in another commit and do some cleanup by removing the `hint` from the labels to reduce confusion in the future -- that can be on a different PR if you'd like to move this forward @ajfarkas

PR #770, opened 2024-04-11
Dmm/fix some policy code

Sgtpluck:
As part of putting together a design strategy for implementing role-based access on the Partner Dashboard, I found a number of ways that the Pundit Policy code could be refactored. Putting that together before starting the bigger process seemed nice! There's definitely more cleanup that could be done, but I wanted to do this while I was thinking about it!

### Description of Changes:
This change:
- renames some misnamed specs
- fixes the user_policy_spec to be actually checking the UserPolicy class
- abstracts out some similar methods in the children Policy classes
- moves a duplicated check for whether a user is admin to the BasePolicy class
- removes children Policy class initializers, as none of them were doing anything that wasn't accomplished in the BasePolicy class

### PR Checklist:

### PR Review Standards: 


app/controllers/service_providers_controller.rb
```diff
@@ -1,5 +1,7 @@
 class ServiceProvidersController < AuthenticatedController
-  before_action :authorize_service_provider
+  before_action -> { authorize ServiceProvider }, only: [:index, :create, :new, :all]
+  before_action -> {
+ authorize service_provider, :member_or_admin? }, only: %i[update edit show destroy]
```
zachmargolis, line 4: spacing got funky here, maybe realign like this? also I'd add parens for clarity
```suggestion
  before_action -> { authorize(service_provider, :member_or_admin?) },
                only: %i[update edit show destroy]
```

app/policies/team_policy.rb
```diff
@@ -1,52 +1,41 @@
 class TeamPolicy < BasePolicy
```
Sgtpluck, line 1: i didn't touch this file much -- it uses our third, elusive role (as defined by email addresses) so it will be updated the most in the actual roles implementation

zachmargolis (approved):

app/controllers/service_providers_controller.rb
Sgtpluck: hahaha this is what rubocop did on its own. bad linter!

nprimak (approved):

app/controllers/service_providers_controller.rb
Sgtpluck: fixed as part of https://github.com/18F/identity-dashboard/pull/770/commits/17dc2014e97cfbfff15583686ae23b9fbadd8b30

PR #772, opened 2024-04-16
LG-13064 HTML parsing bug

ajfarkas:
[LG-13064](https://cm-jira.usa.gov/browse/LG-13064)
[Slack Thread](https://gsa-tts.slack.com/archives/C04RGFTFD70/p1713212455936609)

### Description of Changes:
* Use `t(string_html)` instead of `I18n.t(string_html)`, so that HTML is not escaped to text.
* Updated `span` classes from `text-heavy` to `text-bold`, which is already in our stylesheets and more legible.
* Updated tests to match appropriate text.
* Updated `caniuse-lite`.

### PR Checklist:

### PR Review Standards: 

### DIFF
--- a/app/views/service_providers/_form.html.erb
+++ b/app/views/service_providers/_form.html.erb
@@ -20,13 +20,13 @@
                  input_html: { class: 'usa-input' },
                  label_html: { class: 'usa-input-required'},
                  label: 'App name',
-                 hint: I18n.t('service_provider_form.app_name_html') %>
+                 hint: t('service_provider_form.app_name_html') %>
 
   <%= form.input :friendly_name,
                  input_html: { class: 'usa-input', aria: { invalid: false } },
                  label_html: { class: 'usa-input-required'},
                  label: 'Friendly name',
-                 hint: I18n.t('service_provider_form.friendly_name_html') %>
+                 hint: t('service_provider_form.friendly_name_html') %>
 
 
   <%= form.input :description,
@@ -71,7 +71,7 @@
                  selected: form.object.ial,
                  input_html: { class: 'usa-input usa-select'},
                  label: 'Level of Service',
-                 hint: I18n.t('service_provider_form.identity_assurance_level_html') %>
+                 hint: t('service_provider_form.identity_assurance_level_html') %>
 
   <%= form.input :default_aal,
                  as: :select,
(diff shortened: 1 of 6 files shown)


spec/features/service_providers_spec.rb
```diff
@@ -1,4 +1,7 @@
 require 'rails_helper'
+def StripTags(str)
+  ActionController::Base.helpers.strip_tags(str)
+end
```
zachmargolis, line 4: 1. method names are usually `snake_case`
2. If we want to scope this helper method to just this file, we should define it inside the `feature` block, otherwise it will "leak" to the global namespace and be in all the other files
3. If we want to make this available to all other files, I'd put it in its own file in `spec/support/` somewhere

app/views/service_providers/_form.html.erb
```diff
@@ -20,13 +20,13 @@
                  input_html: { class: 'usa-input' },
                  label_html: { class: 'usa-input-required'},
                  label: 'App name',
-                 hint: I18n.t('service_provider_form.app_name_html') %>
+                 hint: t('service_provider_form.app_name_html') %>
```
zachmargolis, line 23: `t` is basically just an alias to `I18n.t`, I am surprised this made a difference?

Sgtpluck: we were also surprised. (it's also documented [here](https://guides.rubyonrails.org/i18n.html#using-safe-html-translations) in a lil info box)

spec/features/service_providers_spec.rb
Sgtpluck: ➕

app/views/service_providers/_form.html.erb
ajfarkas: Ruby is a language of many mysteries.

nprimak (approved):
LGTM

Sgtpluck (approved):

Jeremy1026 (approved):

spec/features/service_providers_spec.rb
zachmargolis: well one way to shorthand it would be include the `Forwardable ` module and define a delegator, but defining a new method is the easiest to read

Assigning method to a variable in Ruby is possible but trickier due to the lack of distinction between method call + variable access

